PR #3, opened 2019-06-11
Hbase 22567

wchevreuil:
First PR for addMissingRegionsToMeta

### DIFF
--- a/hbase-hbck2/pom.xml
+++ b/hbase-hbck2/pom.xml
@@ -147,11 +147,6 @@
       <artifactId>junit</artifactId>
       <version>4.12</version>
     </dependency>
-    <dependency>
-      <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
-      <version>2.1.0</version>
-    </dependency>
     <!--This is Log4j2!-->
     <!-- https://mvnrepository.com/artifact/org.apache.logging.log4j/log4j-core -->
     <dependency>
@@ -184,6 +179,12 @@
       <groupId>org.apache.hbase</groupId>
       <artifactId>hbase-testing-util</artifactId>
       <version>${hbase.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>2.1.0</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/hbase-hbck2/pom.xml
+++ b/hbase-hbck2/pom.xml
@@ -147,6 +147,11 @@
       <artifactId>junit</artifactId>
       <version>4.12</version>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>2.1.0</version>
+    </dependency>
     <!--This is Log4j2!-->
     <!-- https://mvnrepository.com/artifact/org.apache.logging.log4j/log4j-core -->
     <dependency>
--- a/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
+++ b/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
@@ -178,37 +178,66 @@ public class HBCK2 extends Configured implements Tool {
     final CountDownLatch countDownLatch = new CountDownLatch(tableNames.length);
     final List<String> encodedRegionNames = new ArrayList<>();
     String result = "No regions added.";
-    try(final MetaFixer metaFixer = new MetaFixer()){
-      for(String table : tableNames){
-        executorService.submit(new Runnable() {
-          @Override public void run() {
-            try {
-              System.out.println("running thread for " + table);
-              List<Path> missingRegions = metaFixer.findMissingRegionsInMETA(table);
-              missingRegions.parallelStream().forEach(path -> {
+    try(final MetaFixer metaFixer = new MetaFixer(this.conf)){
+      //reducing number of retries in case disable fails due to namespace table region also missing
+      this.conf.setInt(HConstants.HBASE_CLIENT_RETRIES_NUMBER, 1);
+      try(Connection conn = ConnectionFactory.createConnection(this.conf)) {
+        final Admin admin = conn.getAdmin();
+        for (String table : tableNames) {
+          final TableName tableName = TableName.valueOf(table);
+          if(admin.tableExists(tableName)) {
+            executorService.submit(new Runnable() {
+              @Override public void run() {
                 try {
-                  metaFixer.putRegionInfoFromHdfsInMeta(path);
-                  encodedRegionNames.add(path.getName());
+                  LOG.debug("running thread for {}", table);
+                  List<Path> missingRegions = metaFixer.findMissingRegionsInMETA(table);
+                  try {
+                    admin.disableTable(tableName);
+                  } catch (IOException e) {
+                    LOG.warn("Failed to disable table {}, "
+                        + "is namespace table also missing regions? Continue anyway...",
+                      table, e);
+                  }
+                  missingRegions.stream().forEach(path -> {
+                    try {
+                      metaFixer.putRegionInfoFromHdfsInMeta(path);
+                      encodedRegionNames.add(path.getName());
+                    } catch (Exception e) {
+                      e.printStackTrace();
+                    }
+                  });
+                  try {
+                    admin.enableTable(tableName);
+                  } catch (IOException e) {
+                    LOG.warn("Failed enabling table {}. It might be that namespace table "
+                        + "region is also missing.\n"
+                        + "After this command finishes, please make sure on this table state.",
+                      table, e);
+                  }
                 } catch (Exception e) {
                   e.printStackTrace();
+                } finally {
+                  countDownLatch.countDown();
                 }
-              });
-            } catch (Exception e) {
-              e.printStackTrace();
-            } finally {
-              countDownLatch.countDown();
-            }
+              }
+            });
+          }else{
+            LOG.warn("Table {} does not exist! Skipping...", table);
+            countDownLatch.countDown();
           }
-        });
+        }
+        countDownLatch.await();
+        admin.close();
       }
-      countDownLatch.await();
       StringBuilder finalText = new StringBuilder();
       finalText.append("Regions re-added into Meta: ").append(encodedRegionNames.size());
       if(encodedRegionNames.size()>0){
-        finalText.append("\n\t")
-          .append("You need to run below hbck2 'assigns' command to bring these regions online:")
-          .append("\n\t\t")
-          .append(metaFixer.printHbck2AssignsCommand(encodedRegionNames));
+        finalText.append("\n")
+          .append("WARNING: \n\t")
+          .append(encodedRegionNames.size()).append(" regions were added to ")
+          .append("to META, but these are not yet on Masters cache. \n")
+          .append("You need to restart Masters, then run hbck2 'assigns' command below:\n\t\t")
+          .append(metaFixer.buildHbck2AssignsCommand(encodedRegionNames));
       }
       result = finalText.toString();
     } catch (InterruptedException ie){
@@ -393,19 +422,19 @@ public class HBCK2 extends Configured implements Tool {
       + "\"1\" FAIL code otherwise.");
     writer.println();
     writer.println(" " + ADD_MISSING_REGIONS_IN_META + " <TABLENAME>...");
-    writer.println("   Possible region states: " + Arrays.stream(RegionState.State.values()).
-      map(i -> i.toString()).collect(Collectors.joining(", ")));
-    writer.println("   To be used for scenarios where some regions may be missing in META table,");
+    writer.println("   To be used for scenarios where some regions may be missing in META,");
     writer.println("   but there's still a valid 'regioninfo metadata file on HDFS. ");
     writer.println("   This is a lighter version of 'OfflineMetaRepair tool commonly used for ");
-    writer.println("   similar issues on 1.x release line. This command needs META to be online. ");
-    writer.println("   For each table name passed as parameter, it performs a diff between ");
-    writer.println("   regions available in META, against existing regions dirs on HDFS. ");
-    writer.println("   Then, for region dirs with no matches in META, it reads regioninfo' ");
-    writer.println("   metdata file and re-creates given region in META. ");
-    writer.println("   Regions are re-created in 'CLOSED' state, and are not assigned. ");
-    writer.println("   An hbck2 'assigns' command with all re-inserted regions is printed for ");
-    writer.println("   user convenience.");
+    writer.println("   similar issues on 1.x release line. ");
+    writer.println("   This command needs META to be online. For each table name passed as");
+    writer.println("   parameter, it performs a diff between regions available in META, ");
+    writer.println("   against existing regions dirs on HDFS. Then, for region dirs with ");
+    writer.println("   no matches in META, it reads regioninfo metadata file and ");
+    writer.println("   re-creates given region in META. Regions are re-created in 'CLOSED' ");
+    writer.println("   state at META table only, but not in Masters' cache, and are not ");
+    writer.println("   assigned either. A rolling Masters restart, followed by a ");
+    writer.println("   hbck2 'assigns' command with all re-inserted regions is required. ");
+    writer.println("   This hbck2 'assigns' command is printed for user convenience.");
     writer.println("   WARNING: To avoid potential region overlapping problems due to ongoing ");
     writer.println("   splits, this command disables given tables while re-inserting regions. ");
     writer.println("   An example adding missing regions for tables 'table_1' and 'table_2':");
--- a/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
+++ b/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
@@ -46,6 +46,7 @@ import org.apache.hadoop.hbase.filter.RowFilter;
 import org.apache.hadoop.hbase.filter.SubstringComparator;
 import org.apache.hadoop.hbase.master.RegionState;
 import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.hadoop.hbase.util.Pair;
 import org.apache.hadoop.hbase.util.VersionInfo;
 import org.apache.hadoop.util.Tool;
 import org.apache.hadoop.util.ToolRunner;
@@ -177,125 +178,96 @@ public class HBCK2 extends Configured implements Tool {
     return EXIT_FAILURE;
   }
 
-  Map<String,List<Path>> reportTablesWithMissingRegionsInMeta(String... nameSpaceOrTable)
+  Map<TableName,List<Path>> reportTablesWithMissingRegionsInMeta(String... nameSpaceOrTable)
       throws Exception {
-    final StringBuilder builder = new StringBuilder();
-    Map<String,List<Path>> report;
+    Map<TableName,List<Path>> report;
     try(final MetaFixer metaFixer = new MetaFixer(this.conf)){
       List<String> names = nameSpaceOrTable != null ? Arrays.asList(nameSpaceOrTable) : null;
       report = metaFixer.reportTablesMissingRegions(names);
-      builder.append("Missing Regions for each table:\n\t");
-      report.keySet().stream().forEach(table -> {
-        builder.append(table);
-        if (report.get(table).size()>0){
-          builder.append("->\n\t\t");
-          report.get(table).stream().forEach(region -> builder.append(region.getName())
-            .append(" "));
-        } else {
-          builder.append(" -> No missing regions");
-        }
-        builder.append("\n\t");
-      });
     } catch (Exception e) {
       LOG.error("Error reporting missing regions: ", e);
       throw e;
     }
-    System.out.println(builder);
+    if(LOG.isDebugEnabled()) {
+      LOG.debug(formatMissingRegionsInMetaReport(report));
+    }
     return report;
   }
 
   List<String> addMissingRegionsInMeta(List<Path> regionsPath) throws IOException {
-    List<String> reAddedRegions = new ArrayList<>();
+    List<String> reAddedRegionsEncodedNames = new ArrayList<>();
     try(final MetaFixer metaFixer = new MetaFixer(this.conf)){
       for(Path regionPath : regionsPath){
         metaFixer.putRegionInfoFromHdfsInMeta(regionPath);
-        reAddedRegions.add(regionPath.getName());
+        reAddedRegionsEncodedNames.add(regionPath.getName());
       }
     }
-    return reAddedRegions;
+    return reAddedRegionsEncodedNames;
   }
 
-  int addMissingRegionsInMetaForTables(String... nameSpaceOrTable) throws IOException {
+  Pair<List<String>, List<ExecutionException>> addMissingRegionsInMetaForTables(String...
+      nameSpaceOrTable) {
     ExecutorService executorService = Executors.newFixedThreadPool(
       nameSpaceOrTable.length > Runtime.getRuntime().availableProcessors() ?
         Runtime.getRuntime().availableProcessors() : nameSpaceOrTable.length);
     List<Future<List<String>>> futures = new ArrayList<>(nameSpaceOrTable.length);
-    final List<String> encodedRegionNames = new ArrayList<>();
+    final List<String> readdedRegionNames = new ArrayList<>();
     List<ExecutionException> executionErrors = new ArrayList<>();
-    String resultText = "No regions added.";
-    int result = EXIT_SUCCESS;
     try(final MetaFixer metaFixer = new MetaFixer(this.conf)){
       //reducing number of retries in case disable fails due to namespace table region also missing
       this.conf.setInt(HConstants.HBASE_CLIENT_RETRIES_NUMBER, 1);
-      try(Connection conn = ConnectionFactory.createConnection(this.conf)) {
-        final Admin admin = conn.getAdmin();
-        Map<String,List<Path>> report = this.reportTablesWithMissingRegionsInMeta(nameSpaceOrTable);
-        for (String table : report.keySet()) {
-          final TableName tableName = TableName.valueOf(table);
+      try(Connection conn = ConnectionFactory.createConnection(this.conf);
+        final Admin admin = conn.getAdmin()) {
+        Map<TableName,List<Path>> report = reportTablesWithMissingRegionsInMeta(nameSpaceOrTable);
+        for (TableName tableName : report.keySet()) {
           if(admin.tableExists(tableName)) {
             futures.add(executorService.submit(new Callable<List<String>>() {
               @Override
               public List<String> call() throws Exception {
-                LOG.debug("running thread for {}", table);
+                LOG.debug("running thread for {}", tableName.getNameWithNamespaceInclAsString());
                 try {
                   admin.disableTable(tableName);
                 } catch (IOException e) {
                   LOG.debug("Failed to disable table {}, "
-                      + "is namespace table also missing regions? Continue anyway...", table, e);
+                      + "is namespace table also missing regions? Continue anyway...",
+                    tableName.getNameWithNamespaceInclAsString(), e);
                 }
-                List<String> reAddedRegions = addMissingRegionsInMeta(report.get(table));
+                List<String> reAddedRegions = addMissingRegionsInMeta(report.get(tableName));
                 try {
                   admin.enableTable(tableName);
                 } catch (IOException e) {
                   LOG.debug("Failed enabling table {}. It might be that namespace table "
                       + "region is also missing.\n"
                       + "After this command finishes, please make sure on this table state.",
-                    table, e);
+                    tableName.getNameWithNamespaceInclAsString(), e);
                 }
                 return reAddedRegions;
               }
             }));
-          }else{
-            LOG.warn("Table {} does not exist! Skipping...", table);
+          } else {
+            LOG.warn("Table {} does not exist! Skipping...",
+              tableName.getNameWithNamespaceInclAsString());
           }
         }
         for(Future<List<String>> f : futures){
           try {
-            encodedRegionNames.addAll(f.get());
+            readdedRegionNames.addAll(f.get());
           } catch (ExecutionException e){
             //we want to allow potential running threads to finish, so we collect execution
             //errors and show those later
             executionErrors.add(e);
           }
         }
-        admin.close();
       }
-      final StringBuilder finalText = new StringBuilder();
-      finalText.append("Regions re-added into Meta: ").append(encodedRegionNames.size());
-      if(encodedRegionNames.size()>0){
-        finalText.append("\n")
-          .append("WARNING: \n\t")
-          .append(encodedRegionNames.size()).append(" regions were added ")
-          .append("to META, but these are not yet on Masters cache. \n")
-          .append("You need to restart Masters, then run hbck2 'assigns' command below:\n\t\t")
-          .append(metaFixer.buildHbck2AssignsCommand(encodedRegionNames));
-      }
-      if(executionErrors.size()>0){
-        finalText.append("\n")
-          .append("ERROR: \n\t")
-          .append("There were following errors on at least one table thread:\n");
-        executionErrors.stream().forEach(e -> finalText.append(e.getMessage()).append("\n"));
-        result = EXIT_FAILURE;
-      }
-      resultText = finalText.toString();
     } catch (Exception ie){
       System.out.println("ERROR executing thread: ");
       ie.printStackTrace();
-      return EXIT_FAILURE;
     } finally {
       executorService.shutdown();
     }
-    System.out.println(resultText);
+    Pair<List<String>, List<ExecutionException>> result = new Pair<>();
+    result.setFirst(readdedRegionNames);
+    result.setSecond(executionErrors);
     return result;
   }
 
@@ -669,11 +641,15 @@ public class HBCK2 extends Configured implements Tool {
         usage(options, command + " takes one or more table names.");
         return EXIT_FAILURE;
       }
-      System.out.println(this.addMissingRegionsInMetaForTables(purgeFirst(commands)));
+      Pair<List<String>, List<ExecutionException>> result =
+        addMissingRegionsInMetaForTables(purgeFirst(commands));
+      System.out.println(formatReAddedRegionsMessage(result.getFirst(),result.getSecond()));
       break;
     case REPORT_MISSING_REGIONS_IN_META:
       try {
-        this.reportTablesWithMissingRegionsInMeta(purgeFirst(commands));
+        Map<TableName,List<Path>> report =
+          reportTablesWithMissingRegionsInMeta(purgeFirst(commands));
+        System.out.println(formatMissingRegionsInMetaReport(report));
       } catch (Exception e) {
         return EXIT_FAILURE;
       }
@@ -687,6 +663,51 @@ public class HBCK2 extends Configured implements Tool {
 
   private static String toString(List<?> things) {
     return things.stream().map(i -> i.toString()).collect(Collectors.joining(", "));
+  }
+
+  private String formatMissingRegionsInMetaReport(Map<TableName,List<Path>> report) {
+    final StringBuilder builder = new StringBuilder();
+    builder.append("Missing Regions for each table:\n\t");
+    report.keySet().stream().forEach(table -> {
+      builder.append(table);
+      if (report.get(table).isEmpty()){
+        builder.append("->\n\t\t");
+        report.get(table).stream().forEach(region -> builder.append(region.getName())
+          .append(" "));
+      } else {
+        builder.append(" -> No missing regions");
+      }
+      builder.append("\n\t");
+    });
+    return builder.toString();
+  }
+
+  private String formatReAddedRegionsMessage(List<String> readdedRegionNames,
+    List<ExecutionException> executionErrors) {
+    final StringBuilder finalText = new StringBuilder();
+    finalText.append("Regions re-added into Meta: ").append(readdedRegionNames.size());
+    if(readdedRegionNames.isEmpty()){
+      finalText.append("\n")
+        .append("WARNING: \n\t")
+        .append(readdedRegionNames.size()).append(" regions were added ")
+        .append("to META, but these are not yet on Masters cache. \n")
+        .append("You need to restart Masters, then run hbck2 'assigns' command below:\n\t\t")
+        .append(buildHbck2AssignsCommand(readdedRegionNames));
+    }
+    if(executionErrors.isEmpty()){
+      finalText.append("\n")
+        .append("ERROR: \n\t")
+        .append("There were following errors on at least one table thread:\n");
+      executionErrors.stream().forEach(e -> finalText.append(e.getMessage()).append("\n"));
+    }
+    return finalText.toString();
+  }
+
+  private String buildHbck2AssignsCommand(List<String> regions) {
+    final StringBuilder builder = new StringBuilder();
+    builder.append("assigns ");
+    regions.forEach(region -> builder.append(region).append(" "));
+    return builder.toString();
   }
 
   /**
--- a/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
+++ b/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
@@ -696,7 +696,7 @@ public class HBCK2 extends Configured implements Tool {
     List<ExecutionException> executionErrors) {
     final StringBuilder finalText = new StringBuilder();
     finalText.append("Regions re-added into Meta: ").append(readdedRegionNames.size());
-    if(readdedRegionNames.isEmpty()){
+    if(!readdedRegionNames.isEmpty()){
       finalText.append("\n")
         .append("WARNING: \n\t")
         .append(readdedRegionNames.size()).append(" regions were added ")
@@ -704,7 +704,7 @@ public class HBCK2 extends Configured implements Tool {
         .append("You need to restart Masters, then run hbck2 'assigns' command below:\n\t\t")
         .append(buildHbck2AssignsCommand(readdedRegionNames));
     }
-    if(executionErrors.isEmpty()){
+    if(!executionErrors.isEmpty()){
       finalText.append("\n")
         .append("ERROR: \n\t")
         .append("There were following errors on at least one table thread:\n");
--- a/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
+++ b/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
@@ -441,6 +441,26 @@ public class HBCK2 extends Configured implements Tool {
     writer.println("     $ HBCK2 addMissingRegionsInMeta table_1 table_2");
     writer.println("   Returns hbck2 'assigns' command with all re-inserted regions.");
     writer.println();
+    writer.println(" " + ADD_MISSING_REGIONS_IN_META + " <TABLENAME>...");
+    writer.println("   To be used for scenarios where some regions may be missing in META,");
+    writer.println("   but there's still a valid 'regioninfo metadata file on HDFS. ");
+    writer.println("   This is a lighter version of 'OfflineMetaRepair tool commonly used for ");
+    writer.println("   similar issues on 1.x release line. ");
+    writer.println("   This command needs META to be online. For each table name passed as");
+    writer.println("   parameter, it performs a diff between regions available in META, ");
+    writer.println("   against existing regions dirs on HDFS. Then, for region dirs with ");
+    writer.println("   no matches in META, it reads regioninfo metadata file and ");
+    writer.println("   re-creates given region in META. Regions are re-created in 'CLOSED' ");
+    writer.println("   state at META table only, but not in Masters' cache, and are not ");
+    writer.println("   assigned either. A rolling Masters restart, followed by a ");
+    writer.println("   hbck2 'assigns' command with all re-inserted regions is required. ");
+    writer.println("   This hbck2 'assigns' command is printed for user convenience.");
+    writer.println("   WARNING: To avoid potential region overlapping problems due to ongoing ");
+    writer.println("   splits, this command disables given tables while re-inserting regions. ");
+    writer.println("   An example adding missing regions for tables 'table_1' and 'table_2':");
+    writer.println("     $ HBCK2 addMissingRegionsInMeta table_1 table_2");
+    writer.println("   Returns hbck2 'assigns' command with all re-inserted regions.");
+    writer.println();
 
     writer.close();
     return sw.toString();
--- a/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
+++ b/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
@@ -455,11 +455,15 @@ public class HBCK2 extends Configured implements Tool {
     writer.println("   state at META table only, but not in Masters' cache, and are not ");
     writer.println("   assigned either. A rolling Masters restart, followed by a ");
     writer.println("   hbck2 'assigns' command with all re-inserted regions is required. ");
-    writer.println("   This hbck2 'assigns' command is printed for user convenience.");
+    writer.println("   This hbck2 'assigns' command is printed out after this command ");
+    writer.println("   completes for users subsequent convenience.");
     writer.println("   WARNING: To avoid potential region overlapping problems due to ongoing ");
-    writer.println("   splits, this command disables given tables while re-inserting regions. ");
-    writer.println("   An example adding missing regions for tables 'table_1' and 'table_2':");
-    writer.println("     $ HBCK2 addMissingRegionsInMeta table_1 table_2");
+    writer.println("   splits, this command attempts to disable given tables ");
+    writer.println("   (if it fails, disable is skipped) while re-inserting regions. ");
+    writer.println("   An example adding missing regions for tables 'tbl_1' on default ");
+    writer.println("   namespace, 'tbl_2' on namespace 'n1' and for all tables from ");
+    writer.println("   namespace 'n2': ");
+    writer.println("     $ HBCK2 addMissingRegionsInMeta default:tbl_1 n1:tbl_2 n2 ");
     writer.println("   Returns hbck2 'assigns' command with all re-inserted regions.");
     writer.println();
     writer.println(" " + REPORT_MISSING_REGIONS_IN_META + " <NAMESPACE|"
@@ -467,7 +471,10 @@ public class HBCK2 extends Configured implements Tool {
     writer.println("   To be used in scenarios where some regions may be missing in META,");
     writer.println("   but there's still a valid 'regioninfo metadata file on HDFS. ");
     writer.println("   This is a checking only method, designed for reporting purposes and");
-    writer.println("   doesn't perform any fixes. Operators should react upon the reported.");
+    writer.println("   doesn't perform any fixes, providing a view of which regions (if any) ");
+    writer.println("   would get re-added to meta, grouped by respective table/namespace. ");
+    writer.println("   To effectively re-added regions in meta, "
+      + ADD_MISSING_REGIONS_IN_META_FOR_TABLES + " should be executed. ");
     writer.println("   This command needs META to be online. For each namespace/table passed");
     writer.println("   as parameter, it performs a diff between regions available in META, ");
     writer.println("   against existing regions dirs on HDFS. Region dirs with no matches");
--- a/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
+++ b/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
@@ -64,11 +64,9 @@ import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.EnumSet;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
-import java.util.concurrent.CountDownLatch;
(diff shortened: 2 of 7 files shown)


asfgit:
Refer to this link for build results (access rights to CI server needed): 
https://builds.apache.org/job/PreCommit-HBASE-OPERATOR-TOOLS-Build/4/

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +171,86 @@ int setRegionState(String region, RegionState.State newState)
     return EXIT_FAILURE;
   }
 
+  int addMissingRegionsInMeta(String... tableNames) throws  IOException {
```
joshelser: Could we break this apart into two steps?

1. Report missing regions in meta
2. Add a region to meta

We would be able to compose this higher-level tool then, in something like:
```bash
REGIONS=$(hbase hbck2 report_missing_regions)
for region in REGIONS; do
  hbase hbck2 add_region "region"
done
```

I think this would help keep the complexity of this method down, making it easier to test as a by-product. It would also let an admin inspect the output from the first step (making sure we want to re-create all of those regions, maybe doing some external validation) before trying to re-create them all in one step.

I think the tricky part would be coming up with the "format" to take the output from step1 and pass it to step2 with minimal "massaging".

WDYT about that, Wellington?

busbey: useful for backgound IMHO is git's "plumbing" vs "porcelain":

https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain

analogous here would be "report missing" and "add region" as plumbing level. The output of either need not be user friendly and should focus on being maintainable. (so like encoded region name for the output of report missing would be fine)

something like the wrapper Josh mentions would be a porcelain command, probably taking the encoded regions from report missing and providing a user friendly info (like table, region keys, etc). it'd be useful for it to have a `--dryrun` mode that skipped calling `add_region` so that operators could get the more user friendly version of info about what regions appear to be missing.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +171,86 @@ int setRegionState(String region, RegionState.State newState)
...
+      try(Connection conn = ConnectionFactory.createConnection(this.conf)) {
+        final Admin admin = conn.getAdmin();
+        for (String table : tableNames) {
+          final TableName tableName = TableName.valueOf(table);
+          if(admin.tableExists(tableName)) {
+            executorService.submit(new Runnable() {
+              @Override public void run() {
+                try {
+                  LOG.debug("running thread for {}", table);
+                  List<Path> missingRegions = metaFixer.findMissingRegionsInMETA(table);
```
joshelser: If the configuration is messed up (wrong `hbase.rootdir`, for example), this will bubble up to the `printStackTrace` in the catch. The context that we had this exception thrown will be lost.

Maybe switch the ExecutorService over to using the methods that return Futures to track the completion of all of these tasks? That would also let you get right of the CountDownLatch to track completion.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +171,86 @@ int setRegionState(String region, RegionState.State newState)
...
+      this.conf.setInt(HConstants.HBASE_CLIENT_RETRIES_NUMBER, 1);
+      try(Connection conn = ConnectionFactory.createConnection(this.conf)) {
+        final Admin admin = conn.getAdmin();
+        for (String table : tableNames) {
+          final TableName tableName = TableName.valueOf(table);
+          if(admin.tableExists(tableName)) {
+            executorService.submit(new Runnable() {
+              @Override public void run() {
+                try {
+                  LOG.debug("running thread for {}", table);
```
xcangCRM: Maybe this log line should be info level and more descriptive?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +171,86 @@ int setRegionState(String region, RegionState.State newState)
...
+                  List<Path> missingRegions = metaFixer.findMissingRegionsInMETA(table);
+                  try {
+                    admin.disableTable(tableName);
+                  } catch (IOException e) {
+                    LOG.warn("Failed to disable table {}, "
```
xcangCRM: why is it safe to proceed even failing to disable the table?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +171,86 @@ int setRegionState(String region, RegionState.State newState)
...
+                        + "is namespace table also missing regions? Continue anyway...",
+                      table, e);
+                  }
+                  missingRegions.stream().forEach(path -> {
+                    try {
+                      metaFixer.putRegionInfoFromHdfsInMeta(path);
+                      encodedRegionNames.add(path.getName());
```
xcangCRM: I think for "encodedRegionNames", this variable's name doesn't reflect it's real purpose, mind changing it to clearer one? thanks

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  public List<Path> findMissingRegionsInMETA(String table) throws Exception {
+    final List<Path> missingRegions = new ArrayList<>();
+    final FileStatus[] regionsDirs = getTableRegionsDirs(table);
+    TableName tableName = TableName.valueOf(table);
+    List<RegionInfo> regionInfos = MetaTableAccessor.
+      getTableRegions(this.conn, tableName, false);
+    for(final FileStatus regionDir : regionsDirs){
+      if(!regionDir.getPath().getName().equals(".tabledesc")&&!regionDir.getPath().getName().equals(".tmp")) {
```
xcangCRM: I spotted some minor formatting issues and also can we put ".tabledesc" into a constant? thanks.
nit:
how about moving regionFoundInMeta() into a separate function that we can unit test?

asfgit:
Refer to this link for build results (access rights to CI server needed): 
https://builds.apache.org/job/PreCommit-HBASE-OPERATOR-TOOLS-Build/5/

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: @joshelser @busbey , Sorry for the delay. I think these are all good ideas. Had pushed a commit where I'm trying to implement the plumbing/porcelain approach within 3 methods: _addMissingRegionsInMeta_, _addMissingRegionsInMetaForTables_ and _reportTablesWithMissingRegionsInMeta_. The original porcelain method is now _addMissingRegionsInMetaForTables_. It combines _addMissingRegionsInMeta_ and _reportTablesWithMissingRegionsInMeta_ calls to add regions back in meta. _reportTablesWithMissingRegionsInMeta_ can is also exposed as a CLI method, as a mean to provide a reporting only command for operators. Some additional UTs are still pending, but an initial manual testing on a broke cluster did look promising. Let me know what you guys think about these, while I will add more UTs for those.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Changed to use futures as suggested in the last commit. Let me know if that looks better now, @joshelser .

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Initially thought it could be too verbose, but I guess info should be fine, and this description is indeed not much intuitive. Will address that o next commit.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: We try to disable mostly to make sure we will not be dealing with regions that are transient, say if split/merge is happening while we scan meta. Assumption here is that if disable fails, the table is already somehow offline. That would be the case when even namespace table has missing regions, for example, so we need to proceed with regions re-insertion even if we are not able to disable table.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Sure, maybe something like _readdedRegionNames_ would make more sense?

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
wchevreuil: > I spotted some minor formatting issues and also can we put ".tabledesc" into a constant? 

Indeed, moving it together with ".tmp" into constants. Will be available in next commit.

> how about moving regionFoundInMeta() into a separate function that we can unit test?

I think it's already getting tested by the available tests for _findMissingRegionsInMETA_ method. Since there's not any other client method requiring such logic, I didn't feel it should deserve its own separate method.

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
```diff
@@ -0,0 +1,128 @@
...
+        boolean foundInMeta = regionInfos.stream()
+          .anyMatch(info -> info.getEncodedName().equals(regionDir.getPath().getName()));
+        if (!foundInMeta) {
+          System.out.println(regionDir + "is not in META.");
```
daisukebe: @wchevreuil Wonder if we still have to report this whereas the missing regions are reported inside of `reportTablesWithMissingRegionsInMeta`.

wchevreuil: Hey @daisukebe, this is definitely too verbose and redundant as you pointed. Wondering, though, if it would still be worth having it as a debug/trace message, just in case hbck is performing weird. What would you think?

daisukebe: Yup, making it a debug message makes sense. Thanks!

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +177,128 @@ int setRegionState(String region, RegionState.State newState)
     return EXIT_FAILURE;
   }
 
+  Map<String,List<Path>> reportTablesWithMissingRegionsInMeta(String... nameSpaceOrTable)
```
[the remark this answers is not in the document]
saintstack: What if operator mixes table and namespace in the list passed?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +177,128 @@ int setRegionState(String region, RegionState.State newState)
...
+    final List<String> encodedRegionNames = new ArrayList<>();
+    List<ExecutionException> executionErrors = new ArrayList<>();
+    String resultText = "No regions added.";
+    int result = EXIT_SUCCESS;
+    try(final MetaFixer metaFixer = new MetaFixer(this.conf)){
+      //reducing number of retries in case disable fails due to namespace table region also missing
+      this.conf.setInt(HConstants.HBASE_CLIENT_RETRIES_NUMBER, 1);
+      try(Connection conn = ConnectionFactory.createConnection(this.conf)) {
+        final Admin admin = conn.getAdmin();
+        Map<String,List<Path>> report = this.reportTablesWithMissingRegionsInMeta(nameSpaceOrTable);
```
saintstack: Spacing. No need of the 'this'.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +177,128 @@ int setRegionState(String region, RegionState.State newState)
...
+        for (String table : report.keySet()) {
+          final TableName tableName = TableName.valueOf(table);
```
saintstack: This is small argument in favor of returned map having TableName as key?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +177,128 @@ int setRegionState(String region, RegionState.State newState)
...
+          if(admin.tableExists(tableName)) {
+            futures.add(executorService.submit(new Callable<List<String>>() {
+              @Override
+              public List<String> call() throws Exception {
+                LOG.debug("running thread for {}", table);
+                try {
+                  admin.disableTable(tableName);
```
saintstack: Should this be optional? What if they table has tens of thousands of regions in it? What if you can't disable it because it really damaged?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +177,128 @@ int setRegionState(String region, RegionState.State newState)
...
+                } catch (IOException e) {
+                  LOG.debug("Failed to disable table {}, "
+                      + "is namespace table also missing regions? Continue anyway...", table, e);
```
saintstack: But I suppose if the disable fails and we just keep going, that is probably ok? What does it say when disable fails?

saintstack: If this program dies in the middle of running, will we be able to rerun?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +177,128 @@ int setRegionState(String region, RegionState.State newState)
...
+                } catch (IOException e) {
+                  LOG.debug("Failed enabling table {}. It might be that namespace table "
+                      + "region is also missing.\n"
+                      + "After this command finishes, please make sure on this table state.",
+                    table, e);
+                }
+                return reAddedRegions;
+              }
+            }));
+          }else{
```
saintstack: Spacing

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +177,128 @@ int setRegionState(String region, RegionState.State newState)
...
+    List<Future<List<String>>> futures = new ArrayList<>(nameSpaceOrTable.length);
+    final List<String> encodedRegionNames = new ArrayList<>();
+    List<ExecutionException> executionErrors = new ArrayList<>();
+    String resultText = "No regions added.";
+    int result = EXIT_SUCCESS;
+    try(final MetaFixer metaFixer = new MetaFixer(this.conf)){
+      //reducing number of retries in case disable fails due to namespace table region also missing
+      this.conf.setInt(HConstants.HBASE_CLIENT_RETRIES_NUMBER, 1);
+      try(Connection conn = ConnectionFactory.createConnection(this.conf)) {
+        final Admin admin = conn.getAdmin();
```
saintstack: Don't want this to be inside the try clause? Perhaps it does not implement Closeable.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +177,128 @@ int setRegionState(String region, RegionState.State newState)
...
+            encodedRegionNames.addAll(f.get());
+          } catch (ExecutionException e){
+            //we want to allow potential running threads to finish, so we collect execution
+            //errors and show those later
+            executionErrors.add(e);
+          }
+        }
+        admin.close();
+      }
+      final StringBuilder finalText = new StringBuilder();
```
saintstack: This formatting of output could be a method?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -334,6 +469,51 @@ private static final String getCommandUsage() {
     writer.println("   Returns \"0\" SUCCESS code if it informed region state is changed, "
       + "\"1\" FAIL code otherwise.");
     writer.println();
+    writer.println(" " + ADD_MISSING_REGIONS_IN_META_FOR_TABLES + " <NAMESPACE|"
+      + "NAMESPACE:TABLENAME>...");
+    writer.println("   To be used in scenarios where some regions may be missing in META,");
+    writer.println("   but there's still a valid 'regioninfo metadata file on HDFS. ");
```
saintstack: How we help operator determine this is the case? Doc?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -334,6 +469,51 @@ private static final String getCommandUsage() {
...
+    writer.println("   This is a lighter version of 'OfflineMetaRepair tool commonly used for ");
+    writer.println("   similar issues on 1.x release line. ");
```
saintstack: I should look but is this all that the offlinemetarepair did? Read hdfs .hregioninfo files?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -334,6 +469,51 @@ private static final String getCommandUsage() {
...
+    writer.println("   This command needs META to be online. For each table name passed as");
+    writer.println("   parameter, it performs a diff between regions available in META, ");
+    writer.println("   against existing regions dirs on HDFS. Then, for region dirs with ");
```
saintstack: hmmmm... could there be case where meta is correct and hdfs is wrong? I think so.

saintstack: But we only insert missing regions? If so, then it should be ok.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -334,6 +469,51 @@ private static final String getCommandUsage() {
...
+    writer.println("   no matches in META, it reads regioninfo metadata file and ");
+    writer.println("   re-creates given region in META. Regions are re-created in 'CLOSED' ");
+    writer.println("   state at META table only, but not in Masters' cache, and are not ");
+    writer.println("   assigned either. A rolling Masters restart, followed by a ");
```
saintstack: Why the rolling Masters restart needed?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -334,6 +469,51 @@ private static final String getCommandUsage() {
...
+    writer.println("   hbck2 'assigns' command with all re-inserted regions is required. ");
+    writer.println("   This hbck2 'assigns' command is printed for user convenience.");
```
saintstack: ... printed out after this command completes for users subsequent convenience? 

Suggestion

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -334,6 +469,51 @@ private static final String getCommandUsage() {
...
+    writer.println("   WARNING: To avoid potential region overlapping problems due to ongoing ");
+    writer.println("   splits, this command disables given tables while re-inserting regions. ");
```
saintstack: 'attempts' to disable the table (if it fails, disable is skipped)...

Yeah, disabling the table is safest thing to do.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -334,6 +469,51 @@ private static final String getCommandUsage() {
...
+    writer.println("   An example adding missing regions for tables 'table_1' and 'table_2':");
+    writer.println("     $ HBCK2 addMissingRegionsInMeta table_1 table_2");
+    writer.println("   Returns hbck2 'assigns' command with all re-inserted regions.");
```
saintstack: Good

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -334,6 +469,51 @@ private static final String getCommandUsage() {
...
+    writer.println("   no matches in META, it reads regioninfo metadata file and ");
+    writer.println("   re-creates given region in META. Regions are re-created in 'CLOSED' ");
+    writer.println("   state at META table only, but not in Masters' cache, and are not ");
+    writer.println("   assigned either. A rolling Masters restart, followed by a ");
+    writer.println("   hbck2 'assigns' command with all re-inserted regions is required. ");
+    writer.println("   This hbck2 'assigns' command is printed for user convenience.");
+    writer.println("   WARNING: To avoid potential region overlapping problems due to ongoing ");
+    writer.println("   splits, this command disables given tables while re-inserting regions. ");
+    writer.println("   An example adding missing regions for tables 'table_1' and 'table_2':");
+    writer.println("     $ HBCK2 addMissingRegionsInMeta table_1 table_2");
```
saintstack: Should you add the namespace here too?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -334,6 +469,51 @@ private static final String getCommandUsage() {
...
+    writer.println("   should include the namespace portion, even for tables in the default");
+    writer.println("   namespace, otherwise it will assume as a namespace value.");
+    writer.println("   An example triggering missing regions report for tables 'table_1'");
+    writer.println("   and 'table_2', under default namespace:");
+    writer.println("     $ HBCK2 reportMissingRegionsInMeta default:table_1 default:table_2");
+    writer.println("   An example triggering missing regions report for table 'table_1'");
+    writer.println("   under default namespace, and for all tables from namespace 'ns1':");
+    writer.println("     $ HBCK2 reportMissingRegionsInMeta default:table_1 ns1");
+    writer.println("   Returns list of missing regions for each table passed as parameter, or ");
+    writer.println("   for each table on namespaces specified as parameter.");
```
saintstack: Should this be tied to ADD_MISSING_REGIONS_IN_META_FOR_TABLES more directly? i.e. run this and then run ADD_MISSING_REGIONS_IN_META_FOR_TABLES to effect repair?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -440,51 +620,64 @@ public int run(String[] args) throws IOException {
...
+      System.out.println(toString(unassigns(purgeFirst(commands))));
+      break;
+
+    case SET_REGION_STATE:
+      if (commands.length < 3) {
+        usage(options, command + " takes region encoded name and state arguments: e.g. "
+          + "35f30b0ce922c34bf5c284eff33ba8b3 CLOSING");
+        return EXIT_FAILURE;
+      }
+      return setRegionState(commands[1], RegionState.State.valueOf(commands[2]));
```
saintstack: No 'break' is intentional here?

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
```diff
@@ -0,0 +1,128 @@
...
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+public class MetaFixer implements Closeable {
```
saintstack: Class comment?

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
```diff
@@ -0,0 +1,128 @@
...
+  private static final String HBASE_DATA_DIR = "/data/";
+  private static final String HBASE_DEFAULT_NAMESPACE = "default/";
```
saintstack: Are these two defines available elsewhere? If so, do we have to define them here?

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
```diff
@@ -0,0 +1,128 @@
...
+  private FileSystem fs;
+  private Connection conn;
+  private Configuration config;
```
saintstack: final?

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+public class MetaFixer implements Closeable {
+  private static final String HBASE_DATA_DIR = "/data/";
+  private static final String HBASE_DEFAULT_NAMESPACE = "default/";
+  private FileSystem fs;
```
saintstack: final?

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
```diff
@@ -0,0 +1,128 @@
...
+  private Connection conn;
```
saintstack: final?

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
```diff
@@ -0,0 +1,128 @@
...
+  /*Initially defined for test only purposes */
+  MetaFixer(Configuration configuration, Connection connection, FileSystem fileSystem){
+    this.config = configuration;
+    this.conn = connection;
+    this.fs = fileSystem;
+  }
+
+  private FileStatus[] getTableRegionsDirs(String table) throws Exception {
+    String hbaseRoot = this.config.get("hbase.rootdir");
+    String tableRootDir = hbaseRoot + (table.indexOf(":") < 0 ?
```
saintstack: Nothing in TableName class to do this for you? TableName is on your classpath?

saintstack: i.e. the namespace messing.

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
```diff
@@ -0,0 +1,128 @@
...
+        HBASE_DATA_DIR + table.trim().replaceAll(":", "/"));
+    return fs.listStatus(new Path(tableRootDir));
+  }
+
+  public Map<String,List<Path>> reportTablesMissingRegions(final List<String> namespacesOrTables)
+      throws IOException {
+    final Map<String,List<Path>> result = new HashMap<>();
+    List<TableName> tableNames = MetaTableAccessor.getTableStates(this.conn).keySet().stream()
+      .filter(tableName -> {
+        if(namespacesOrTables==null || namespacesOrTables.size()==0){
```
saintstack: isEmpty instead of size() == 0

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
```diff
@@ -0,0 +1,128 @@
...
+          return true;
+        } else {
+          Optional<String> findings = namespacesOrTables.stream().filter(
+            name -> (name.indexOf(":") > 0) ?
```
saintstack: Nothing in TableName to do this already?

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
```diff
@@ -0,0 +1,128 @@
...
+        result.put(tableName.getNameWithNamespaceInclAsString(),
+          findMissingRegionsInMETA(tableName.getNameWithNamespaceInclAsString()));
+      } catch (Exception e) {
+        e.printStackTrace();
+      }
+    });
+    return result;
+  }
+
+  public List<Path> findMissingRegionsInMETA(String table) throws Exception {
```
saintstack: These methods need to be public?

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
saintstack: tabledesc not already in a constant over in hbase?

saintstack: Spacing ... need spaces around operators as we have in rest of codebase.

Yeah, is the .tmp defined already too?

saintstack: Is there not a filitering mechanism elsewhere to rule out these exceptions? Could it be reused?

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
```diff
@@ -0,0 +1,128 @@
...
+          System.out.println(regionDir + "is not in META.");
+          missingRegions.add(regionDir.getPath());
+        }
+      }
+    }
+    return missingRegions;
+  }
+
+  public void putRegionInfoFromHdfsInMeta(Path region) throws IOException {
+    RegionInfo info = HRegionFileSystem.loadRegionInfoFileContent(fs, region);
```
saintstack: Should there be a check we don't overlap with an existing region?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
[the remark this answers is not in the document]
wchevreuil: It would still list missing regions specific for each namespace/table passed as `parameter.` Even if we pass a namespace and then a table within that namespace, output would show the missing regions grouped. For example, say namespace _ns1_ has two tables _tbl1_ and _tbl2_, and each of these two tables has a missing region _r1_ and _r2_, respectively. Then calling `reportTablesWithMissingRegionsInMeta ns1 ns1:tbl1`, should print:
`ns1 -> r1 r2`
`ns1:tbl1 -> r1`

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +177,128 @@ int setRegionState(String region, RegionState.State newState)
     return EXIT_FAILURE;
   }
 
+  Map<String,List<Path>> reportTablesWithMissingRegionsInMeta(String... nameSpaceOrTable)
+      throws Exception {
+    final StringBuilder builder = new StringBuilder();
+    Map<String,List<Path>> report;
```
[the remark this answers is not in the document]
wchevreuil: Makes sense, addressing it on next PR commit.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +177,128 @@ int setRegionState(String region, RegionState.State newState)
...
+    try(final MetaFixer metaFixer = new MetaFixer(this.conf)){
+      List<String> names = nameSpaceOrTable != null ? Arrays.asList(nameSpaceOrTable) : null;
+      report = metaFixer.reportTablesMissingRegions(names);
+      builder.append("Missing Regions for each table:\n\t");
+      report.keySet().stream().forEach(table -> {
+        builder.append(table);
+        if (report.get(table).size()>0){
```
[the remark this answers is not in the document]
wchevreuil: Definitely! Addressing it on next commit.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +177,128 @@ int setRegionState(String region, RegionState.State newState)
...
+        } else {
+          builder.append(" -> No missing regions");
+        }
+        builder.append("\n\t");
+      });
+    } catch (Exception e) {
+      LOG.error("Error reporting missing regions: ", e);
+      throw e;
+    }
+    System.out.println(builder);
```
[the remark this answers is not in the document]
wchevreuil: Agreed, addressing on next commit.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +177,128 @@ int setRegionState(String region, RegionState.State newState)
...
+    return report;
+  }
+
+  List<String> addMissingRegionsInMeta(List<Path> regionsPath) throws IOException {
```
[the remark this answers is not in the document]
wchevreuil: Yep, the region encoded names for regions that were missing and got re-added in meta. Should I rename that variable, or maybe add javadocs?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Well spotted, had missed that! Addressing it on next commit.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Noted. Addressing on next commit.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Yep, no need for that with TableName as key. Moved this line out after applying related change. Addressing it on next commit.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Maybe we don't really need to disable the table, but decided to add this as an extra safety measure. On the cases I had seen so far, affected tables were not online, or not being accessed, but what if we end up running this for a table receiving many puts, to the point where region splits happen frequently? We could end up in a race condition situation between this command and the split proc while checking/updating meta. Does split proc create daughter regions dirs before daughter regions info are in meta? Or does catalog janitor clears out split parents from meta before it removes its directories? Those are situations that could potentially lead to meta inconsistencies, then.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: > But I suppose if the disable fails and we just keep going, that is probably ok? What does it say when disable fails?
There are some scenarios where disable will fail for sure, but we still should proceed, for instance, when namespace table is also affected (and on the cases we faced so far, this was happening). Earlier versions of this command we used with some customers was just printing the exception, then continuing, but it was confusing operators, leading them to think the program had actually failed, when it had not. So, I had decided to leave it as debugging log.

> If this program dies in the middle of running, will we be able to rerun?
It doesn't really rely on any state, so if a given run listed, say 100 missing regions, then went re-adding those and crashed after re-inserting 50 regions, next run would simply list the remaining 50 and iterate through those. I guess this would be a good test case scenario, let me add it to my backlog of work to be done after the most trivial ones.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Noted. Addressing on next commit.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Yeah, makes more sense. Addressing it on next commit.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Good point. Maybe another section on the _readme_?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: It does more complex stuffs, like validating region ranges among existing hdfs region dirs. Also, it sidelines and recreates meta from the scratch, then insert table and region info (read from hdfs) into the newly created meta table.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: The most critical features I can think of here are Splits and Merges, but I believe CatalogJanitor would move retired regions dirs to archive when it deletes given region from meta, right? Retired regions still in meta with region dirs wouldn't be a problem, because as long as these still have entries on meta, it wouldn't be marked as missing. Would you think of any valid scenario where meta doesn't have a given region, but this region dir is still on hdfs (non archived)? If so, then I think it could be a problem because this command would recreate this region info in meta.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: It's the penalty for doing things without consent from Master :)...

We are performing client puts of region infos into meta. While it does insert the region back into meta, Active _Master.AssignmentManager.RegionStateStore_ never gets updated about the new region, so any attempt to assign these re-added regions will fail, because AM doesn't know these regions. I couldn't find, currently, any master rpc available method that could trigger a meta _reload_.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Thanks! Will be addressed in next commit.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Done, will be in next commit.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Yep, addressing it on next commit.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Thanks!

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: This came out of previous suggestions to this PR [here](https://github.com/apache/hbase-operator-tools/pull/3#discussion_r292981293). It's basically a type of dry-run for operators double check on what is going to be re-added to meta. Let me emphasise that on this command description.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Yeah, because we are already returning output of _setRegionState_ method, a _break_ would be unreachable. Maybe we should change _setRegionState_ case handling to become consistent with other methods? Probably on a separate jira.

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
wchevreuil: Done.

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
wchevreuil: Done.

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
wchevreuil: Done.

asfgit:
Refer to this link for build results (access rights to CI server needed): 
https://builds.apache.org/job/PreCommit-HBASE-OPERATOR-TOOLS-Build/6/

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
wchevreuil: Yep, coming on next commit.

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
wchevreuil: Not exactly in this format, but there's HConstants.BASE_NAMESPACE_DIR and NamespaceDescriptor.DEFAULT_NAMESPACE_NAME_STR, which could be used here. After applying changes to address next comment, though,

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
wchevreuil: Indeed, much simpler by creating a _TableName_ instance, then using _getNameWithNamespaceInclAsString:_
`TableName.valueOf(table).getNameWithNamespaceInclAsString()
        .replaceAll(":", "/");`

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
wchevreuil: Coming on next commit.

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
wchevreuil: Here _name_ comes from the list of method params, which are simple strings. I need this because params can be either in the format _namespace_ or _namespace:table_, so I need to distinguish between the formats, in order to decide which filter logic to apply: 1) case _namespace_, then use _TableName.getNamespaceAsString_; 2) case _namespace:table_, then use _TableName.getNameWithNamespaceInclAsString_; 
Does it make sense?

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
wchevreuil: Not anymore, originally this was being called from _HBCK2_ class directly, but now is only used internally by _MetaFixer_. Is it ok to leave it as package private? It's conveniently being used by _TestMetaFixer_.

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
wchevreuil: Coming on next commit.

asfgit:
Refer to this link for build results (access rights to CI server needed): 
https://builds.apache.org/job/PreCommit-HBASE-OPERATOR-TOOLS-Build/7/

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
wchevreuil: Good point. My first idea was to just leave it break, and operator would need to take further actions, such as merge overlapping regions. We could add additional check, as a mean to warn operator that extra merge would be required later. 

Any other thoughts?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
joshelser: Yup, the use of Futures makes that much cleaner. Will have to dig in about the issues of hbase:namespace being messed up :)

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
busbey: We should add one that hbck2 can call?

hbase-hbck2/pom.xml
```diff
@@ -147,6 +147,11 @@
       <artifactId>junit</artifactId>
       <version>4.12</version>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>2.1.0</version>
```
busbey: should be in test scope

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +178,99 @@ int setRegionState(String region, RegionState.State newState)
...
+          } else {
+            LOG.warn("Table {} does not exist! Skipping...",
+              tableName.getNameWithNamespaceInclAsString());
+          }
+        }
+        for(Future<List<String>> f : futures){
+          try {
+            readdedRegionNames.addAll(f.get());
+          } catch (ExecutionException e){
+            //we want to allow potential running threads to finish, so we collect execution
```
busbey: maybe show the in-progress error at DEBUG in case one of these running threads hangs indefinitely?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
busbey: weird. I thought I sent this comment in response a "I don't see a service that can tell a running master to reload meta"

busbey: found it. must have mess up hitting buttons on mobile

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
busbey: We should add one that hbck2 can call?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -334,6 +441,58 @@ private static final String getCommandUsage() {
     writer.println("   Returns \"0\" SUCCESS code if it informed region state is changed, "
       + "\"1\" FAIL code otherwise.");
     writer.println();
+    writer.println(" " + ADD_MISSING_REGIONS_IN_META_FOR_TABLES + " <NAMESPACE|"
+      + "NAMESPACE:TABLENAME>...");
+    writer.println("   To be used in scenarios where some regions may be missing in META,");
```
busbey: this cli help needs to go into the README as well. Could some of the text on what to do _after_ running the command go into a README under a new *Fixing* section?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: For the _caught_ errors, am already debugging it, but _uncaught_ errors would interrupt the thread and be shown later whilst we go through this _executionErrors_ list. Would you think we still need to debug any potential _ExecutionException_ here?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -496,6 +672,51 @@ private static String toString(List<?> things) {
...
+    finalText.append("Regions re-added into Meta: ").append(readdedRegionNames.size());
+    if(readdedRegionNames.isEmpty()){
+      finalText.append("\n")
+        .append("WARNING: \n\t")
+        .append(readdedRegionNames.size()).append(" regions were added ")
+        .append("to META, but these are not yet on Masters cache. \n")
+        .append("You need to restart Masters, then run hbck2 'assigns' command below:\n\t\t")
+        .append(buildHbck2AssignsCommand(readdedRegionNames));
+    }
+    if(executionErrors.isEmpty()){
```
busbey: should be if the list is not empty

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
busbey: yeah, one of the futures could refuse to be interrupted and then we'd never get to the point of the nicely formatted output when we go through the executionErrors.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -164,6 +178,99 @@ int setRegionState(String region, RegionState.State newState)
...
+            readdedRegionNames.addAll(f.get());
+          } catch (ExecutionException e){
+            //we want to allow potential running threads to finish, so we collect execution
+            //errors and show those later
+            executionErrors.add(e);
+          }
+        }
+      }
+    } catch (Exception ie){
+      System.out.println("ERROR executing thread: ");
```
busbey: use the LOG instead of System.out? (should be System.err if we're skipping the logging system)

hbase-hbck2/pom.xml
wchevreuil: Addressed on PR commit d40cfb1b22450b03e6d2d4ff38e5e6ea4cb5857e.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Addressed on PR commit d40cfb1b22450b03e6d2d4ff38e5e6ea4cb5857e.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Addressed in PR commit d40cfb1b22450b03e6d2d4ff38e5e6ea4cb5857e.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Yep, this revealed itself as a bug when I last try to use this command. Addressed in PR commit d40cfb1b22450b03e6d2d4ff38e5e6ea4cb5857e.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Addressed in PR commit d40cfb1.

wchevreuil:
Had pushed a new commit d40cfb1, addressing latest suggestions and adding more UTs.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Good idea! I will submit a separate jira for this.

wchevreuil: Opened HBASE-22707 for this. Decided to embed it on current _MasterRpcServices.assigns_ command, instead of defining a whole new RPC method for meta reload.

saintstack:
Looking at this again, because I'm looking at how to have hbck2 plug holes. This is a hole plugger but only for the case where there is a region in HDFS that has been dropped. This should run first. Lets get it in (smile). There are outstanding comments. Maybe take a look?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Since HBASE-22707 is already committed, there's no need for the extra master restart. I will update readme and help instructions on a next commit.

busbey: it's still needed if the HBase you're fixing doesn't have HBASE-22707 right?

busbey:
I'd also like to get this landed as a part of getting a released artifact out for hbase-operator-tools soon. I'll keep an eye out for the update, but if I miss it and you're looking for reviews please ping me.

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
wchevreuil: > tabledesc not already in a constant over in hbase?
I could only find a package private constant in FSTableDescriptors, so couldn't reuse it here.

wchevreuil: > Spacing ... need spaces around operators as we have in rest of codebase.
Addressed on current PR version.

wchevreuil: > Yeah, is the .tmp defined already too?

Yep, found several constants defined for it, decided to use _HConstants.HBASE_TEMP_DIRECTORY_ on current PR version.

wchevreuil: > Is there not a filitering mechanism elsewhere to rule out these exceptions? Could it be reused?

I'm not aware of any, but maybe there is. Any ideas on where about in the code to to look for? Tried review some of the hbase-server util package classes, but no luck.

hbase-hbck2/src/main/java/org/apache/hbase/hbck2/meta/MetaFixer.java
wchevreuil: > tabledesc not already in a constant over in hbase?

I could only find a package private constant in FSTableDescriptors, so couldn't reuse it here.

> Spacing ... need spaces around operators as we have in rest of codebase.

Addressed on current PR version.

> Yeah, is the .tmp defined already too?

Yep, found several constants defined for it, decided to use HConstants.HBASE_TEMP_DIRECTORY on current PR version.

> Is there not a filitering mechanism elsewhere to rule out these exceptions? Could it be reused?

I'm not aware of any, but maybe there is. Any ideas on where about in the code to to look for? Tried review some of the hbase-server util package classes, but no luck.

wchevreuil:
Had started a new PR after rebasing these changes with latest master version:
https://github.com/apache/hbase-operator-tools/pull/18

Am closing this one.

PR #4, opened 2019-07-10
HBASE-22675 Use the commons-cli from hbase-thirdparty

joshelser:
Should work with any hbase-thirdparty >=2.1.0 and increases
compatibility scope for HBCK2 to more HBase releases.

### DIFF
--- a/hbase-hbck2/pom.xml
+++ b/hbase-hbck2/pom.xml
@@ -32,6 +32,7 @@
   <name>Apache HBase - HBCK2</name>
   <description>HBCK for HBase 2+</description>
   <properties>
+    <hbase-thirdparty.version>2.2.1</hbase-thirdparty.version>
     <log4j2.version>2.11.1</log4j2.version>
   </properties>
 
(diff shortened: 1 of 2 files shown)


hbase-hbck2/pom.xml
```diff
@@ -32,6 +32,7 @@
   <name>Apache HBase - HBCK2</name>
   <description>HBCK for HBase 2+</description>
   <properties>
+    <hbase-thirdparty.version>2.2.1</hbase-thirdparty.version>
```
saintstack: Do you know if this stops hbck from working against hbase2.0.x or 2.1.x? Probably not. I can try it in a while... asking in meantime.

joshelser: I don't. It _shouldn't_ have an impact as this is just compile-time. Should've tested myself :)

ankitsinghal: hbck2 worked fine for me with 2.0.x(atleast) without this change as well. 👍

saintstack: Great.

wchevreuil: Tested compile it with default version and 2.1, then ran assigns/unassigns on both master and branch-2.1 clusters, and it worked fine on both. LGTM+1.

joshelser:
Thanks, folks. Let me resolve this.

If this continues to be a problem, we can start shading stuff into HBCK2 directly. My hope is that we'll have enough from 3rdparty that we can rely on instead.

PR #8, opened 2019-07-13
HBASE-22687 Added HBase Checkstyle configuration to the project

HorizonNet:


HorizonNet:
Added the Apache license header for the property file, because the build complains about it. Because of that, I also changed how the property file is read (reading a property from it instead of reading the whole file).

HorizonNet:
I created HBASE-22691 to move the suppression file to a different location.

PR #11, opened 2019-07-16
HBASE-21322 Add a scheduleServerCrashProcedure() API to HbckService

saintstack:
Signed-off-by: Wellington Chevreuil <wchevreuil@apache.org>

### DIFF
--- a/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
+++ b/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
@@ -49,7 +49,6 @@ import org.apache.hadoop.hbase.filter.RowFilter;
 import org.apache.hadoop.hbase.filter.SubstringComparator;
 import org.apache.hadoop.hbase.master.RegionState;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.HBaseProtos;
-
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
--- a/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
+++ b/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
@@ -50,9 +50,6 @@ import org.apache.hadoop.hbase.filter.SubstringComparator;
 import org.apache.hadoop.hbase.master.RegionState;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.HBaseProtos;
 
-
-import org.apache.hadoop.util.Tool;
-import org.apache.hadoop.util.ToolRunner;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -75,7 +72,7 @@ import org.apache.hbase.thirdparty.org.apache.commons.cli.ParseException;
 // + Doc how we just take pointer to zk ensemble... If want to do more exotic config. on client,
 // then add a hbase-site.xml onto CLASSPATH for this tool to pick up.
 // + Add --version
-public class HBCK2 extends Configured implements Tool {
+public class HBCK2 extends Configured implements org.apache.hadoop.util.Tool {
   private static final Logger LOG = LogManager.getLogger(HBCK2.class);
   private static final int EXIT_SUCCESS = 0;
   static final int EXIT_FAILURE = 1;
@@ -597,7 +594,7 @@ public class HBCK2 extends Configured implements Tool {
 
   public static void main(String [] args) throws Exception {
     Configuration conf = HBaseConfiguration.create();
-    int errCode = ToolRunner.run(new HBCK2(conf), args);
+    int errCode = org.apache.hadoop.util.ToolRunner.run(new HBCK2(conf), args);
     if (errCode != 0) {
       System.exit(errCode);
     }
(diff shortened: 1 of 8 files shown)


saintstack:
All about import order of imports... Punting on doing any more here. Was going to merge...

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -88,6 +91,7 @@
   private static final String FILESYSTEM = "filesystem";
   private static final String VERSION = "version";
   private static final String SET_REGION_STATE = "setRegionState";
+  private static final String SCHEDULE_SCP = "serverCrashProcedures";
```
infraio: How about using scheduleSCP as the cmd?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -254,6 +258,24 @@ int setRegionState(String region, RegionState.State newState)
...
   }
 
+  List<Long> scheduleSCP(String[] args) throws IOException {
+    List<HBaseProtos.ServerName> serverNames = new ArrayList<>();
+    for (String serverName: args) {
+      serverNames.add(parseServerName(serverName));
+    }
+    try (ClusterConnection c = (ClusterConnection)ConnectionFactory.createConnection(getConf())) {
+      try (Hbck hbck = c.getHbck()) {
+        return hbck.scheduleServerCrashProcedure(serverNames);
```
infraio: What will happened if user use this for 2.1.1? Should we add a version check here?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -339,15 +360,22 @@ private static String getCommandUsage() {
     writer.println("     $ HBCK2 setTableState users ENABLED");
     writer.println("   Returns whatever the previous table state was.");
     writer.println();
+    writer.println(" " + SCHEDULE_SCP + " <SERVERNAME>...");
+    writer.println("   Schedule ServerCrashProcedure(SCP) for list of RegionServers. Format");
```
infraio: Schedule ServerCrashProcedure(SCP) for one regionserver?

saintstack: It actual can do a list of servers.... it submits them all in the one lump (see test).

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
saintstack: Thought about that and figured it could get messy having a complicated matrix of what version support what methods. An old idea is that we'd ask the server what 'features' it supports and then we'd fail if wanted-api was missing. This would be a bit tough in this case since we can't change the server-side after the fact. Also sometimes we want to schedule procedures w/o checking server version/features -- i.e. during master startup, a check for version comes back with a PleaseHoldException.

That said, let me make a start. Can do it here w/ this feature.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
saintstack: Would rather not use acronym. It is explained in help but seems cryptic as command. scheduleServerCrashProcedures is too long. What you reckon?

saintstack: Hmm. I'd rather call it 'crash'. The other procedure commands are short.

infraio: I prefer to use a verb as command...

busbey: `scheduleRecovery` ?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
busbey: the version checking stuff is causing me a bunch of heartburn. If I can help on the feature negotiation stuff between hbck2 and the cluster please point me in a direction.

saintstack: I don't think it is going to be too bad. Patch coming.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
saintstack: Called it scheduleRecoveries (to match other commands which are plural to reflect they can do bulk)

saintstack:
Ok. Going to stop messing w/ this checkstyle thing. Its my favorite, import order. Makes no sense. Reviews please!

hbase-hbck2/src/main/java/org/apache/hbase/Version.java
```diff
@@ -0,0 +1,115 @@
...
+          Integer.parseInt(thresholdVersionComponents[PATCH])) {
+        supported = true;
+      }
+      break;
+    }
+    return supported || excessiveMajor || excessiveMinor;
+  }
+
+  /**
+   * Copied from hbase VersionInfo.
```
infraio: Why don't use hbase VersionInfo?

saintstack: The getComponents in private in VersionInfo so have to copy it.

VI doesn't have the check we do here.

saintstack:
@infraio  Thanks.

Sorry for confusion over in HBASE-22673. Lets go w/ this as is for now. Can switch to new API later.

PR #13, opened 2019-07-19
HBASE-22717 [HBCK2] Expose replication fixes from hbck1

saintstack:


saintstack:
Merged small change that exposes what used to be in place in hbck1.

PR #14, opened 2019-08-14
HBASE-22843 [HBCK2] Fix HBCK2 after HBASE-22777 & HBASE-22758

jatsakthi:


jatsakthi:
I think it's because of the 2.1.6-SNAPSHOT in pom.xml, that the mvninstall failed in precommit.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKConstants.java
```diff
@@ -0,0 +1,22 @@
...
+   * Merge qualifier prefix.
+   * We used to only allow two regions merge; mergeA and mergeB.
+   * Now we allow many to merge. Each region to merge will be referenced
+   * in a column whose qualifier starts with this define.
+   */
+  public static final String MERGE_QUALIFIER_PREFIX_STR = "merge";
+
+  public static final byte [] MERGE_QUALIFIER_PREFIX =
+      Bytes.toBytes(MERGE_QUALIFIER_PREFIX_STR);
+}
```
saintstack: We have Constants in hbase but this pulling out constants to be in their own file is an anti-pattern (Constants is a mistake). Constants should be beside where they are used. If shared, then put in base or first mention class or the class that looks best like a 'home' for the constant.

FYI, new files need license.

I see here you are pulling over the new regex facility.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBaseFsck.java
```diff
@@ -3924,13 +3924,14 @@ public boolean visit(Result result) throws IOException {
...
             }
           }
-          PairOfSameType<RegionInfo> mergeRegions = MetaTableAccessor.getMergeRegions(result);
-          for (RegionInfo mergeRegion : new RegionInfo[] {
-              mergeRegions.getFirst(), mergeRegions.getSecond() }) {
-            if (mergeRegion != null) {
-              // This region is already been merged
-              HbckInfo hbInfo = getOrCreateInfo(mergeRegion.getEncodedName());
-              hbInfo.setMerged(true);
+          List<RegionInfo> mergeParents = HBCKMetaTableAccessor.getMergeRegions(result.rawCells());
```
saintstack: This part is good.

pom.xml
```diff
@@ -123,7 +123,7 @@
     <compileSource>1.8</compileSource>
     <java.min.version>${compileSource}</java.min.version>
     <maven.min.version>3.3.3</maven.min.version>
-    <hbase.version>2.1.2</hbase.version>
+    <hbase.version>2.1.6-SNAPSHOT</hbase.version>
```
saintstack: Why this? Makes me worried we'll end up w/ a dependency on new facility when we are trying to keep HBCK 'universal'.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKPrivateCellUtil.java
```diff
@@ -0,0 +1,32 @@
+package org.apache.hbase.hbck1;
```
saintstack: Needs license.

Why in this package? It is not hbck1 facility, right?

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKPrivateCellUtil.java
```diff
@@ -0,0 +1,32 @@
...
+import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.util.ByteBufferUtils;
+import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.yetus.audience.InterfaceAudience;
+
+/**
+ * hbck's local version of PrivateCellUtil from the hbase repo
+ */
+@InterfaceAudience.Private
+final class HBCKPrivateCellUtil {
```
saintstack: PrivateCellUtil is an internal serverside thing. It can be plain CellUtil here. In fact, this facility probably best belongs in CellUtil. It looks harmless except for the bit about ByteBufferExtendedCell. That is internal to hbase. Shouldn't be used outside of core hbase -- i.e. not by HBCK2.

HBCK2 version could just copy the qualifier from the Cell ... clone it and then compare against that. It does not have to be performant.

So, rename this class as CellUtil I think and do not use internal hbase classes that are private like ByteBufferExtendedCell. Move class to top level since it named CellUtil (no need to put into a 'util' subpackage). 

You might also consider just adding this single method to your new HBCKMetaTableAccessor class rather than create a new standalone class. COuld be private internal method.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKConstants.java
saintstack: Where is this used? If used in one class only, add the define to that one class.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * hbck's local version of the MetaTableAccessor from the hbase repo
+ */
+@InterfaceAudience.Private
+class HBCKMetaTableAccessor {
+
+  private static final Logger LOG = LoggerFactory.getLogger(HBCKMetaTableAccessor.class);
+  private static final Logger HBCKMETALOG = LoggerFactory.getLogger("org.apache.hadoop.hbase.META");
+
+  public static List<RegionInfo> getMergeRegions(Cell[] cells) {
```
saintstack: For methods only used by this class make them private. Especially over here in HBCK2. They are only used by this class?

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
```diff
@@ -0,0 +1,134 @@
...
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * hbck's local version of the MetaTableAccessor from the hbase repo
+ */
+@InterfaceAudience.Private
+class HBCKMetaTableAccessor {
```
saintstack: Ok. So this makes sense I think -- dup'ing the notion that there is one way to go to meta from HBCK2. Good, you have it package private.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
saintstack: You might in comment point at the original version of this method... say copied from ... blah, blah.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
```diff
@@ -0,0 +1,134 @@
...
+    // Check to see if has family and that qualifier starts with the merge qualifier 'merge'
+    return CellUtil.matchingFamily(cell, HConstants.CATALOG_FAMILY) &&
+        HBCKPrivateCellUtil.qualifierStartsWith(cell, HBCKConstants.MERGE_QUALIFIER_PREFIX);
+  }
+
+  /**
+   * Returns the column family used for meta columns.
+   * @return HConstants.CATALOG_FAMILY.
+   */
+  public static byte[] getCatalogFamily() {
```
saintstack: Keep private. Maybe remove. Its a bit silly... could just use constant.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBaseFsckRepair.java
```diff
@@ -200,6 +200,6 @@ public static HRegion createHDFSRegionDir(Configuration conf,
    */
   public static void removeParentInMeta(Configuration conf, RegionInfo hri) throws IOException {
     Connection conn = ConnectionFactory.createConnection(conf);
-    MetaTableAccessor.deleteRegion(conn, hri);
```
saintstack: Is this only place where we use the internal MTA? If so, good.

saintstack: Oh, yeah, the method name changed.  Good.

saintstack: So, old HBCK2 won't be able to talk to newer hbase. Ok. We need to update doc on HBCK2 to say use newest version. I think that is ok. HBCK2 should nearly always be newere than the hbase it is probling. Means that in the field there will be some interesting failures if old HBCK2... but should be self-explanatory (method does not exist!)

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
```diff
@@ -0,0 +1,134 @@
+package org.apache.hbase.hbck1;
```
saintstack: Yeah, wrong package.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
```diff
@@ -0,0 +1,134 @@
...
+        regionsToMerge.add(ri);
+      }
+    }
+    return regionsToMerge;
+  }
+
+  private static boolean isMergeQualifierPrefix(Cell cell) {
+    // Check to see if has family and that qualifier starts with the merge qualifier 'merge'
+    return CellUtil.matchingFamily(cell, HConstants.CATALOG_FAMILY) &&
+        HBCKPrivateCellUtil.qualifierStartsWith(cell, HBCKConstants.MERGE_QUALIFIER_PREFIX);
```
saintstack: YEah, just move this method in here rather than create a whole new class.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
```diff
@@ -0,0 +1,134 @@
...
+  /**
+   * Delete the passed <code>d</code> from the <code>hbase:meta</code> table.
+   * @param connection connection we're using
+   * @param d Delete to add to hbase:meta
+   */
+  private static void deleteFromMetaTable(final Connection connection, final Delete d)
+      throws IOException {
+    List<Delete> dels = new ArrayList<>(1);
+    dels.add(d);
+    deleteFromMetaTable(connection, dels);
```
saintstack: Not you I know but maybe better to write as

deleteFromMetaTable(connection, Arrays.asList(d));

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  /**
+   * Callers should call close on the returned {@link Table} instance.
+   * @param connection connection we're using to access Meta
+   * @return An {@link Table} for <code>hbase:meta</code>
+   */
+  public static Table getMetaHTable(final Connection connection)
```
saintstack: s/getMetaHTable/getMetaTable/ or just getTable since this is for accessing one table only.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
```diff
@@ -0,0 +1,134 @@
...
+      return;
+    }
+    // Logging each mutation in separate line makes it easier to see diff between them visually
+    // because of common starting indentation.
+    for (Mutation mutation : mutations) {
+      debugLogMutation(mutation);
+    }
+  }
+
+  private static void debugLogMutation(Mutation p) throws IOException {
```
saintstack: Whats this about? We need it?

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
saintstack: Or do we even need this in this little class?

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
```diff
@@ -0,0 +1,134 @@
...
+  /**
+   * Delete the passed <code>RegionInfo</code> from the <code>hbase:meta</code> table.
+   * @param connection connection we're using
+   * @param regionInfo the regionInfo to delete from the meta table
+   * @throws IOException
+   */
+  public static void deleteRegionInfo(Connection connection, RegionInfo regionInfo)
+      throws IOException {
+    Delete delete = new Delete(regionInfo.getRegionName());
+    delete.addFamily(getCatalogFamily(), HConstants.LATEST_TIMESTAMP);
```
saintstack: If only one use of getCatalogFamily method, just use the constant and drop method.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKConstants.java
jatsakthi: Removed the class and defined it within the class.

pom.xml
jatsakthi: I guess, we would have to eventually take it up to 2.1.6 considering the fact that we want to call hbck.fixMeta() (which was added server side) which was added in 2.1.6?

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKPrivateCellUtil.java
jatsakthi: Moved the method inside class & made a private one.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
jatsakthi: Done.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
jatsakthi: Umm, after moving the class out of this package, couldn't keep it package private anymore.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
jatsakthi: Done. :)

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBaseFsckRepair.java
jatsakthi: HBCKMTA is being used in 2 places as of now. 

`List<RegionInfo> mergeParents = HBCKMetaTableAccessor.getMergeRegions(result.rawCells());` in hbck1.HbaseFsck

`HBCKMetaTableAccessor.deleteRegionInfo(conn, hri);` in hbck1.HbaseFsckRepair

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
jatsakthi: Moved it out of the hbck1 package.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
jatsakthi: Done.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
jatsakthi: Consolidated.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
jatsakthi: Done.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
jatsakthi: Removed this logging. Added a normal debug log, we anyways are only doing a single delete mutation.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBCKMetaTableAccessor.java
jatsakthi: Done.

hbase-hbck2/src/main/java/org/apache/hbase/HBCKMetaTableAccessor.java
```diff
@@ -39,12 +39,13 @@
  * hbck's local version of the MetaTableAccessor from the hbase repo
  */
 @InterfaceAudience.Private
-class HBCKMetaTableAccessor {
+public class HBCKMetaTableAccessor {
```
saintstack: It can stay package private? Or is it that class in hbck1 refers to it? If so, I suppose you have to make this change yeah.

pom.xml
saintstack: True. Lets do that in another issue when we add the fixMeta call?

hbase-hbck2/src/main/java/org/apache/hbase/HBCKMetaTableAccessor.java
```diff
@@ -0,0 +1,133 @@
...
+   * @param left       the cell with which we need to match the qualifier
+   * @param startsWith the serialized keyvalue format byte[]
+   * @return true if the qualifier have same staring characters, false otherwise
+   */
+  private static boolean qualifierStartsWith(final Cell left, final byte[] startsWith) {
+    if (startsWith == null || startsWith.length == 0) {
+      throw new IllegalArgumentException("Cannot pass an empty startsWith");
+    }
+    return Bytes
+        .equals(left.getQualifierArray(), left.getQualifierOffset(), startsWith.length, startsWith,
```
saintstack: This does copy if ByteBuffer backed? Just wondering.

hbase-hbck2/src/main/java/org/apache/hbase/HBCKMetaTableAccessor.java
```diff
@@ -0,0 +1,133 @@
...
+   * @param d          Delete to add to hbase:meta
+   */
+  private static void deleteFromMetaTable(final Connection connection, final Delete d)
+      throws IOException {
+    if (connection == null) {
+      throw new NullPointerException("No connection");
+    } else if (connection.isClosed()) {
+      throw new IOException("connection is closed");
+    }
+    try (Table t = connection.getTable(TableName.META_TABLE_NAME)) {
```
saintstack: Good.

hbase-hbck2/src/main/java/org/apache/hbase/HBCKMetaTableAccessor.java
jatsakthi: Yes, Stack. It's being used by hbck1.HbaseFsck & hbck1.HbaseFsckRepair.

hbase-hbck2/src/main/java/org/apache/hbase/HBCKMetaTableAccessor.java
jatsakthi: Let me double-check this one.

jatsakthi:
So, now there are 2 ways to access the Meta table from hbck2. MTA & HBCKMTA. There are lots of instances of MTA usage. Are we planning to slowly move out of it to have just HBCKMTA and refractor the MTA usages with HBCKMTA? @saintstack

hbase-hbck2/src/main/java/org/apache/hbase/HBCKMetaTableAccessor.java
jatsakthi: Yeah so left.getQualifierArray takes of that I guess. It does a CellUtil.cloneQualifier(this)  for byteBuffer backed cells, which in turn calls copyQualifierTo(), which takes care of bytebuffer copying by ByteBufferUtils.copyFromBufferToArray(). Let me know if it doesn't sound correct Stack.

saintstack:
I think we move over to new MTA when issue using the old. At least for now.

hbase-hbck2/src/main/java/org/apache/hbase/HBCKMetaTableAccessor.java
saintstack: This sounds good. WIll be slower but that ok for fixup tool. Thanks for checking.

saintstack:
Why the failure?

jatsakthi:
One of the checkstyle issues apparently. let me fix it.

jatsakthi:
All tests passed. Let me merge this one. Thanks @saintstack

PR #16, opened 2019-08-20
HBASE-22825 [HBCK2] Add a client-side to hbase-operator-tools that ca…

jatsakthi:
…n exploit fixMeta added in server side

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -123,7 +123,7 @@
     <compileSource>1.8</compileSource>
     <java.min.version>${compileSource}</java.min.version>
     <maven.min.version>3.3.3</maven.min.version>
-    <hbase.version>2.1.2</hbase.version>
+    <hbase.version>2.1.6-SNAPSHOT</hbase.version>
     <maven.compiler.version>3.6.1</maven.compiler.version>
     <surefire.version>2.21.0</surefire.version>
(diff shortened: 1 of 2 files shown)


jatsakthi:
Haven't added UTs, as it would be very similar to what we have in TestMetaFixer. Let me know if we would still like it to be present.

busbey:
Do we require the change to bump the HBase 2.1 version to 2.1.6 specifically? If so we'll have to wait for the release.

jatsakthi:
Yes Sean, for availing the fixMeta api we need to.

pom.xml
```diff
@@ -123,7 +123,7 @@
     <compileSource>1.8</compileSource>
     <java.min.version>${compileSource}</java.min.version>
     <maven.min.version>3.3.3</maven.min.version>
-    <hbase.version>2.1.2</hbase.version>
+    <hbase.version>2.1.6-SNAPSHOT</hbase.version>
```
busbey: 2.1.6 is out now.

saintstack:
Let me close this one and make a new one w/ @jatsakthi 's patch amended referring to released 2.1.6.

busbey:
Sounds good. Next time you should be able to push a commit into this PR with the amendment you want, fyi.

PR #20, opened 2019-08-30
HBASE-22952 HBCK2 replication command is incompatible with 2.0.x

petersomogyi:
Adds supported versions to the `replication` command.

### DIFF
--- a/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
+++ b/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
@@ -566,7 +566,7 @@ public class HBCK2 extends Configured implements org.apache.hadoop.util.Tool {
 
       case REPLICATION:
         try (ClusterConnection connection = connect()) {
-          checkHBCKSupport(connection, command);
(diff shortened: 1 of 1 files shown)


hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -566,7 +566,7 @@ private int doCommandLine(CommandLine commandLine, Options options) throws IOExc
 
       case REPLICATION:
         try (ClusterConnection connection = connect()) {
-          checkHBCKSupport(connection, command);
+          checkHBCKSupport(connection, command, "2.1.1", "2.2.0", "3.0.0");
```
busbey: shouldn't it be 2.1.0 instead of 2.1.1.?

petersomogyi: The first supported 2.1 release is 2.1.1.
MINIMUM_HBCK2_VERSION: https://github.com/apache/hbase-operator-tools/blob/master/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java#L92

PR #23, opened 2019-09-06
Fix invalid function call (from usage to showErrorMessage)

jfrabaute:


jfrabaute:
Thanks for the quick merge!

(Yes, I checked the error log of the jenkins build and it was complaining about the code without the PR applied. Weird.)

PR #24, opened 2019-09-06
HBASE-22984 [HBCK2] HBCKMetaTableAccessor.deleteFromMetaTable throwin…

wchevreuil:
Adding UT and solution for the error.


wchevreuil:
We should get this resolve before I can commit HBASE-22567, otherwise some of the UTs added there will fail hbck2 build.

PR #25, opened 2019-09-09
HBASE-22997 Move to SLF4J

petersomogyi:
* Remove org.apache.logging.log4j dependency
* Add org.slf4j dependency
* Modify Logger initialization
* Remove unused Logger from TestHBCKMetaTableAccessor

### DIFF
--- a/hbase-hbck2/pom.xml
+++ b/hbase-hbck2/pom.xml
@@ -161,10 +161,9 @@
       <artifactId>log4j-core</artifactId>
       <version>${log4j2.version}</version>
     </dependency>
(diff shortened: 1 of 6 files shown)


petersomogyi:
I just realized that the current log configuration is in XML format, probably it needs to be migrated to a properties file.

petersomogyi:
Updated PR based on @busbey 's comments over Jira.

petersomogyi:
> Logs that come out look ok?

Yes, log messages are the same as before and the `-d` debug flag works as well.

PR #30, opened 2019-09-12
HBASE-23002 [HBCK2/hbase-operator-tools] Create an assembly that buil…

saintstack:
…ds an hbase-operator-tools tgz

Add an hbase-assembly that is just for creating assemblies.
Also changed groupid from org.apache.hbase to
org.apache.hbase.operator.tools so lands into a different area in
repository -- less crowded, easier to differentiate from core
hbase.

Creates a bin assembly (src is made elsewhere in release script).
Adds a first-cut at RELEASENOTES and CHANGES (release script will
fix up what is here based off what is in JIRA to make it suit
the 1.0.0RC0)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,0 +1,90 @@
+# HBASE Operator Tools Changelog
+
+<!---
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Be careful doing manual edits in this file. Do not change format
+# of release header or remove the below marker. This file is generated.
+# DO NOT REMOVE THIS MARKER; FOR INTERPOLATING CHANGES!-->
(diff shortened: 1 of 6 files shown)


CHANGES.md
```diff
@@ -0,0 +1,90 @@
+# HBASE Operator Tools Changelog
```
petersomogyi: Why the CHANGES file included with the assembly module's introduction?

RELEASENOTES.md
```diff
@@ -0,0 +1,1228 @@
+# RELEASENOTES
```
petersomogyi: Same with RELEASENOTES

hbase-operator-tools-assembly/src/main/assembly/bin.xml
```diff
@@ -0,0 +1,97 @@
...
+    <file>
+      <source>${project.parent.basedir}/CHANGES.md</source>
+      <outputDirectory>.</outputDirectory>
+      <destName>CHANGES.md</destName>
+      <lineEnding>unix</lineEnding>
+    </file>
+    <file>
+      <source>${project.parent.basedir}/README.md</source>
+      <outputDirectory>.</outputDirectory>
+      <destName>README.md</destName>
```
petersomogyi: README.md is also included in `moduleSets`. Are both of these lines are required?

CHANGES.md
saintstack: We want CHANGES and RELEASENOTES included in the binary tgz, right? Need placeholders.

RELEASENOTES.md
saintstack: Ditto

hbase-operator-tools-assembly/src/main/assembly/bin.xml
saintstack: One is the high-level README for the repo. The other is for hbase-hbck.

saintstack:
Thanks for review Peter.

saintstack:
>  I assume the content of changes and releasenotes will be updated on release creation time.

Script fills in Release notes and Changes... yes sir. Thanks for review.

PR #32, opened 2019-09-13
HBASE-23021 [hbase-operator-tools] README edits in prep for release

saintstack:


hbase-hbck2/README.md
```diff
@@ -519,38 +548,50 @@ command needs to be executed later, so copy and save it for convenience.
...
+```
 
 To assign the namespace table region, you cannot use the shell. If you use the shell, it will fail with a `PleaseHoldException` because the master is not yet up (it is waiting for the namepace table to come online before it declares itself ‘up’). You have to use the `HBCK2` _assigns_ command. To assign, you will need the namespace encoded name. It shows in the log quoted above: i.e. _725a0fe6c2c869d3d0a9ed82bfa80fa3_ in this case. You will also have to pass the -skip command to ‘skip’ the master version check (without it, your `HBCK2` invocation will also elicit the above `PleaseHoldException` because the master is not yet up). Here is an example adding an assign of the namespace table:
-```$ HBASE_CLASSPATH_PREFIX=~/checkouts/hbase-operator-tools/hbase-hbck2/target/hbase-hbck2-1.0.0-SNAPSHOT.jar ./bin/hbase org.apache.hbase.HBCK2 -skip assigns 725a0fe6c2c869d3d0a9ed82bfa80fa3```
+```
+$ HBASE_CLASSPATH_PREFIX=~/checkouts/hbase-operator-tools/hbase-hbck2/target/hbase-hbck2-1.0.0-SNAPSHOT.jar ./bin/hbase org.apache.hbase.HBCK2 -skip assigns 725a0fe6c2c869d3d0a9ed82bfa80fa3
```
busbey: We back to this instead of `hbase hbck -j` ?

saintstack: Thats wrong. Thats old doc. Let me fix. Thanks.

saintstack:
New push addresses what @busbey  found (and checkstyle complaint).

busbey:
For some reason merging this is failing without an explanation. Will check again when at a laptop.

saintstack:
It is failing because I just pushed locally instead.... Thanks for review @busbey

PR #33, opened 2019-09-15
HBASE-23025 Do not quote GPG command

petersomogyi:


petersomogyi:
> Should the GPG variable be an array instead?

Probably that would also work. I'm not that proficient in bash to judge but if you recommend array here I can take a look.

busbey:
It'll only be needed if we need to have either a GPG command or one of the args with a space or something the shell would parse as a glob.

Probably fine to do that later as part of a general shellcheck cleanup.

petersomogyi:
I don't think we have any of these currently.

PR #36, opened 2019-09-16
HBASE-23033 Do not run git-commit-id-plugin when .git is missing

petersomogyi:


saintstack:
Confirmed this PR fixes issue.

PR #40, opened 2019-10-03
HBASE-23112 [hbase-operator-tools] fixMeta in hbck2 is porcelain, in …

saintstack:
…hbck1 it was plumbing; fix

Removed 'porcelain vs plubming' analogy. Its fine when we are talking
amongst ourselves but it confuses describing hbck2, a jumble of tools
doing various things.


hbase-hbck2/README.md
```diff
@@ -20,32 +20,33 @@
...
+similarly named across the two versions. See the next section for how the tools
+differ.
 
 ## Philosophy
-_HBCK2_ performs a single discrete 'fix' task each time it is run. It does not presume
+_HBCK2_ performs a single, discrete task each time it is run. It does not presume
 a tool can analyze all about the running cluster and then repair 'all problems' found as
-_hbck1_ used suggest. _HBCK2_ is a tool that is more in the vein of
-[`plumbing` than `porcelain`](https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain).
+_hbck1_ used suggest.
```
daisukebe: Well, if I understood correctly, `fixMeta` fixes multiple holes and overlaps being reported by CatalogJanitor, right? Is it still `a single, discrete task`?

hbase-hbck2/README.md
```diff
@@ -159,7 +160,9 @@ Command:
 
  fixMeta
    Do a server-side fixing of bad or inconsistent state in hbase:meta.
-   Repairs 'holes' and 'overlaps' in hbase:meta.
+   Critical that hbase:meta is first made healthy before making any
+   other repairs. Fix 'holes', 'overlaps', etc. Is NOT same as the
+   _hbck1_ command named similarily.
    SEE ALSO: reportMissingRegionsInMeta
```
daisukebe: Can we also describe how to recognize the issue which requires to involve this command?

hbase-hbck2/README.md
saintstack: True that it can do a few holes/overlaps behind the scene. True also that fixMeta is a single task.

What should we do here? In old days hbck1 had one command to for fixing holes, another for overlaps. We don't want to expose all the little steps needed to fix meta.... I've seen case where you fix holes and overlaps and meta is still borked. hbck1 used to also run multiple times because it would only do adjacents. The hbck2 fixMeta does better than this using the multi-merge. Its more likely to fix all in the one pass.  We'd like to get fixMeta to a place where it can run once and take care of it all. May take some time.

hbase-hbck2/README.md
saintstack: We can, at least in newer hbases where the holes and overlaps are listed in the UI hbck report. Let me add a note.

hbase-hbck2/README.md
daisukebe: Ahh, ok. No worries, your wording makes sense. Thanks!

hbase-hbck2/README.md
saintstack: Hmmm... This usage note needed a good bit more. See what you think of the latest @daisukebe ? Thanks boss.

daisukebe: Thanks Stack! Reviewed the latest one and I'm still wondering if the Master UI is the only one way to recognize the inconsistency. Is there an equivalent tool developed in the command line?

> Fixes 'holes', 'overlaps', etc

Does it make sense to you to change the sentence to `Fixes 'holes', 'overlaps' in HDFS, etc`?

saintstack: There is no reporting apart from Master UI in hbase2. No equivalent tool. We'd have to write one for those older releases. It would be different though because the report would not be generated by the catalogjanitor but by a tool reading hbase:meta remote running the code we recently added to catalogjanitor.

So, the fixMeta fixes holes and overlaps in meta. It does not fix holes and overlaps in HDFS -- intentionally. Idea is you'd fix meta then if any orphan data in HDFS, you'd ask HBase to adopt it using the bulk load tool to load it up (The filesystem command would note the orphan data/regions).

Shout if I should add more.

daisukebe: > It does not fix holes and overlaps in HDFS -- intentionally.

Ohk, per my reading of the code, https://github.com/apache/hbase/blob/branch-2.2/hbase-server/src/main/java/org/apache/hadoop/hbase/master/MetaFixer.java#L83-L108, it appears to be creating an empty region in the filesystem and then put the entry to meta. I was apparently misunderstanding?

saintstack: You are right in that it creates an empty region to match the hole or overlap found in hbase:meta.

What it does not do is look around in hdfs to try and match hdfs content w/ the hole that is being plugged or moving data around in hdfs to fill the overlap.

I think talking of HDFS side-effects when describing fixMeta will give the wrong impression, that we will be doing reconcile w/ HDFS content which we are explicitly not doing in this command.

Not sure how better to say this boss.

daisukebe: Ah ohk. To me, it looks the `fixHoles` and `fixOverlaps` methods are doing what `-fixHdfsHoles` and `-fixHdfsOverlaps` used to do, respectively, in hbck1, hence I supposed it'd be better to document it that `fixMeta` touches and modifies HDFS. However, you are expecting more than that when mentioning `Fixes 'holes', 'overlaps' in HDFS, etc`, correct?

saintstack: Let me add a note that empty dirs are created in HDFS. Makes sense.

saintstack: Here is what I changed it to in coming patch...

"   Do a server-side fix of bad or inconsistent state in hbase:meta.
   Available in hbase 2.2.1/2.1.6 or newer versions. Master UI has
   matching, new 'HBCK Report' tab that dumps reports generated by
   most recent run of _catalogjanitor_ and a new 'HBCK Chore'. It
   is critical that hbase:meta first be made healthy before making
   any other repairs. Fixes 'holes', 'overlaps', etc., creating
   (empty) region directories in HDFS to match regions added to
   hbase:meta. Command is NOT the same as the old _hbck1_ command
   named similarily. Works against the reports generated by the last
   catalog_janitor and hbck chore runs. If nothing to fix, run is a
   noop. Otherwise, if 'HBCK Report' UI reports problems, a run of
   fixMeta will clear up hbase:meta issues. See 'HBase HBCK' UI
   for how to generate new report.
   SEE ALSO: reportMissingRegionsInMeta"

daisukebe: Thanks for making the change! Looks more user friendly. +1 to the latest patch.

saintstack:
Thanks for the reivews @daisukebe

PR #45, opened 2019-12-05
HBASE-23371 [HBCK2] Provide client side method for removing ghost regions in meta.

wchevreuil:
Figured there were many similarities with the already existing _addFsRegionsMissingInMeta_ and _reportMissingRegionsInMeta_, so had worked on some refactoring in order to avoid code duplication.

### DIFF
--- a/hbase-hbck2/README.md
+++ b/hbase-hbck2/README.md
@@ -174,12 +174,45 @@ Command:
    for how to generate new report.
    SEE ALSO: reportMissingRegionsInMeta
 
+ removeExtraRegionsFromMeta <NAMESPACE|NAMESPACE:TABLENAME>...
+   To be used when regions present on hbase:meta, but with no related
+   directories on the file system. Needs hbase:meta
+   to be online. For each table name passed as parameter, performs diff
+   between regions available in hbase:meta and region dirs on the given
+   file system, removing extra regions from meta with no matching directory.
+   An example removing extra regions for tables 'tbl_1' in the default
+   namespace, 'tbl_2' in namespace 'n1' and for all tables from
+   namespace 'n2':
+     $ HBCK2 removeExtraRegionsFromMeta default:tbl_1 n1:tbl_2 n2
+   SEE ALSO: reportExtraRegionsInMeta
+   SEE ALSO: addFsRegionsMissingInMeta
+   SEE ALSO: fixMeta
+
  replication [OPTIONS] [<TABLENAME>...]
    Options:
     -f, --fix    fix any replication issues found.
    Looks for undeleted replication queues and deletes them if passed the
    '--fix' option. Pass a table name to check for replication barrier and
    purge if '--fix'.
+
+ reportExtraRegionsInMeta <NAMESPACE|NAMESPACE:TABLENAME>...
+   To be used when regions present on hbase:meta, but with no related
+   directories on the file system. Needs hbase:meta to be online.
+   For each table name passed as parameter, performs diff
+   between regions available in hbase:meta and region dirs on the given
+   file system. This is a CHECK only method,
+   designed for reporting purposes and doesn't perform any fixes.
+   It provides a view of which regions (if any) would get removed from meta,
+   grouped by respective table/namespace. To effectively
+   remove regions from meta, run removeExtraRegionsFromMeta.
+   An example triggering extra regions report for tables 'table_1'
+   and 'table_2', under default namespace:
+     $ HBCK2 reportExtraRegionsInMeta default:table_1 default:table_2
+   An example triggering missing regions execute for table 'table_1'
+   under default namespace, and for all tables from namespace 'ns1':
+     $ HBCK2 reportExtraRegionsInMeta default:table_1 ns1
+   Returns list of extra regions for each table passed as parameter, or
+   for each table on namespaces specified as parameter.
 
  reportMissingRegionsInMeta <NAMESPACE|NAMESPACE:TABLENAME>...
    To be used when regions missing from hbase:meta but directories
@@ -527,6 +560,18 @@ operational. In such situations, problem can be addressed with the Master online
 using the _addFsRegionsMissingInMeta_ command in _HBCK2_. This command is less disruptive to
 hbase than a full hbase:meta rebuild covered later, and it can be used even for
 recovering the _namespace_ table region.
+
+### Extra Regions in hbase:meta region/table restore/rebuild
+
+There can also be situations where table regions have been removed file system, but still
+have related entries on hbase:meta table. This may happen due to problems on splitting, manual
+operation mistakes (like deleting/moving the region dir manually), or even meta info data loss
+issues such as HBASE-21843.
+
+Such problem can be addressed with the Master online, using the _removeExtraRegionsFromMeta_
+command in _HBCK2_. This command is less disruptive to hbase than a full hbase:meta rebuild
+covered later. Also useful when this happens on versions that don't support _fixMeta_ hbck2 option
+(any prior to "2.0.6", "2.1.6", "2.2.1", "2.3.0","3.0.0").
 
 #### Online hbase:meta rebuild recipe
 
--- a/hbase-hbck2/README.md
+++ b/hbase-hbck2/README.md
@@ -148,6 +148,29 @@ Command:
    to finish parent and children. This is SLOW, and dangerous so use
    selectively. Does not always work.
 
+ extraRegionsInMeta <NAMESPACE|NAMESPACE:TABLENAME>...
+   Options:
+    -f, --fix    fix meta by removing all extra regions found.
+   Reports regions present on hbase:meta, but with no related
+   directories on the file system. Needs hbase:meta to be online.
+   For each table name passed as parameter, performs diff
+   between regions available in hbase:meta and region dirs on the given
+   file system. Extra regions would get deleted from Meta
+   if passed the --fix option.
+   NOTE: Before deciding on use the "--fix" option, it's worth check if
+   reported extra regions are overlapping with existing valid regions.
+   If so, then "extraRegionsInMeta --fix" is indeed the optimal solution.
+   Otherwise, "assigns" command is the simpler solution, as it recreates
+   regions dirs in the filesystem, if not existing.
+   An example triggering extra regions report for tables 'table_1'
+   and 'table_2', under default namespace:
+     $ HBCK2 extraRegionsInMeta default:table_1 default:table_2
+   An example triggering missing regions execute for table 'table_1'
+   under default namespace, and for all tables from namespace 'ns1':
+     $ HBCK2 extraRegionsInMeta default:table_1 ns1
+   Returns list of extra regions for each table passed as parameter, or
+   for each table on namespaces specified as parameter.
+
  filesystem [OPTIONS] [<TABLENAME>...]
    Options:
     -f, --fix    sideline corrupt hfiles, bad links, and references.
@@ -174,45 +197,12 @@ Command:
    for how to generate new report.
    SEE ALSO: reportMissingRegionsInMeta
 
- removeExtraRegionsFromMeta <NAMESPACE|NAMESPACE:TABLENAME>...
-   To be used when regions present on hbase:meta, but with no related
-   directories on the file system. Needs hbase:meta
-   to be online. For each table name passed as parameter, performs diff
-   between regions available in hbase:meta and region dirs on the given
-   file system, removing extra regions from meta with no matching directory.
-   An example removing extra regions for tables 'tbl_1' in the default
-   namespace, 'tbl_2' in namespace 'n1' and for all tables from
-   namespace 'n2':
-     $ HBCK2 removeExtraRegionsFromMeta default:tbl_1 n1:tbl_2 n2
-   SEE ALSO: reportExtraRegionsInMeta
-   SEE ALSO: addFsRegionsMissingInMeta
-   SEE ALSO: fixMeta
-
  replication [OPTIONS] [<TABLENAME>...]
    Options:
     -f, --fix    fix any replication issues found.
    Looks for undeleted replication queues and deletes them if passed the
    '--fix' option. Pass a table name to check for replication barrier and
    purge if '--fix'.
-
- reportExtraRegionsInMeta <NAMESPACE|NAMESPACE:TABLENAME>...
-   To be used when regions present on hbase:meta, but with no related
-   directories on the file system. Needs hbase:meta to be online.
-   For each table name passed as parameter, performs diff
-   between regions available in hbase:meta and region dirs on the given
-   file system. This is a CHECK only method,
-   designed for reporting purposes and doesn't perform any fixes.
-   It provides a view of which regions (if any) would get removed from meta,
-   grouped by respective table/namespace. To effectively
-   remove regions from meta, run removeExtraRegionsFromMeta.
-   An example triggering extra regions report for tables 'table_1'
-   and 'table_2', under default namespace:
-     $ HBCK2 reportExtraRegionsInMeta default:table_1 default:table_2
-   An example triggering missing regions execute for table 'table_1'
-   under default namespace, and for all tables from namespace 'ns1':
-     $ HBCK2 reportExtraRegionsInMeta default:table_1 ns1
-   Returns list of extra regions for each table passed as parameter, or
-   for each table on namespaces specified as parameter.
 
  reportMissingRegionsInMeta <NAMESPACE|NAMESPACE:TABLENAME>...
    To be used when regions missing from hbase:meta but directories
--- a/hbase-hbck2/README.md
+++ b/hbase-hbck2/README.md
@@ -165,7 +165,7 @@ Command:
    An example triggering extra regions report for tables 'table_1'
    and 'table_2', under default namespace:
      $ HBCK2 extraRegionsInMeta default:table_1 default:table_2
-   An example triggering missing regions execute for table 'table_1'
+   An example triggering extra regions report for table 'table_1'
    under default namespace, and for all tables from namespace 'ns1':
      $ HBCK2 extraRegionsInMeta default:table_1 ns1
    Returns list of extra regions for each table passed as parameter, or
--- a/hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
+++ b/hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
@@ -30,6 +30,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import org.apache.hadoop.conf.Configuration;
@@ -51,7 +52,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * This class implements the inner works required for checking and recover regions that wrongly
+ * This class implements the inner works required for checking and recovering regions that wrongly
  * went missing in META, or are left present in META but with no equivalent FS dir.
  * Normally HBCK2 fix options rely on Master self-contained information to recover/fix
  * inconsistencies, but this an exception case where META table is in a broken state.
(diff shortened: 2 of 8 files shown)


hbase-hbck2/README.md
```diff
@@ -174,13 +174,46 @@ Command:
    for how to generate new report.
    SEE ALSO: reportMissingRegionsInMeta
 
+ removeExtraRegionsFromMeta <NAMESPACE|NAMESPACE:TABLENAME>...
+   To be used when regions present on hbase:meta, but with no related
+   directories on the file system. Needs hbase:meta
+   to be online. For each table name passed as parameter, performs diff
+   between regions available in hbase:meta and region dirs on the given
+   file system, removing extra regions from meta with no matching directory.
```
saintstack: Removing regions from meta? Favoring the filesystem view?

hbase-hbck2/README.md
```diff
@@ -174,13 +174,46 @@ Command:
...
+   SEE ALSO: fixMeta
+
  replication [OPTIONS] [<TABLENAME>...]
    Options:
     -f, --fix    fix any replication issues found.
    Looks for undeleted replication queues and deletes them if passed the
    '--fix' option. Pass a table name to check for replication barrier and
    purge if '--fix'.
 
+ reportExtraRegionsInMeta <NAMESPACE|NAMESPACE:TABLENAME>...
```
saintstack: You don't want to add one method and then add a --fix flag for case of where you want the command to actually perform fixes? (filesystem does this -- reporting if no flags and then fixing if a --fix passed).

saintstack: This cuts down some on the number of commands operator sees.

daisukebe:
Hi @wchevreuil, couple of comments based on my test. 

Unless the region is overlapped with other regions in meta, uses should use `assign '<region>'` command as a viable action to take because it generates a new empty region directory appropriately via `org.apache.hadoop.hbase.regionserver.HRegionFileSystem#writeRegionInfoFileContent`. 

Otherwise, in a case where the region is overlapped with others in meta, we should remove the region's entry from meta as assuming it does not have valid data in HDFS.

Thoughts?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -99,8 +97,10 @@
 
   private static final String ADD_MISSING_REGIONS_IN_META_FOR_TABLES =
     "addFsRegionsMissingInMeta";
-  private static final String ADD_MISSING_REGIONS_IN_META = "addMissingRegionsInMeta";
   private static final String REPORT_MISSING_REGIONS_IN_META = "reportMissingRegionsInMeta";
+  static final String REMOVE_EXTRA_REGIONS_IN_META_FROM_TABLES =
+    "removeExtraRegionsFromMeta";
+  static final String REPORT_EXTRA_REGIONS_IN_META = "reportExtraRegionsInMeta";
```
daisukebe: People in the real world, including me, are still using hbck1 as a trusted report to deal with inconsistencies. Good to have this functionality in HBCK2 too!

hbase-hbck2/README.md
wchevreuil: Yeah (see Daisuke's comments next). This is similar to the previous _addFsRegionsMissingInMeta_ command, where we assume the filesystem to be the source of truth. Given the "ghost" region in meta is useless without a related dir in the filesystem, here deletion would have no risk of dataloss.

hbase-hbck2/README.md
wchevreuil: Yep, makes sense, let me refactor it accordingly. Coming on next commit.

wchevreuil: Yep, makes sense, let me refactor it accordingly. Coming on next commit.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: > Unless the region is overlapped with other regions in meta, uses should use assign '<region>' command as a viable action to take because it generates a new empty region directory appropriately via org.apache.hadoop.hbase.regionserver.HRegionFileSystem#writeRegionInfoFileContent.

Thanks for pointing this out, @daisukebe ! Let me add this valuable insight to the command instructions and "recipes" section of the README.

> People in the real world, including me, are still using hbck1 as a trusted report to deal with inconsistencies

Yeah, hbck2 is not targeted to be a reporting tool, but as part of work for both this new command and the _addFsRegionsMissingInMeta_ is to figure out which regions these need to fix, had then decided to expose the _checking_ logic as a reporting command that can be executed on its own.

wchevreuil:
New commit pushed to this PR. Changes:
1) Made a single command "extraRegionsInMeta" with a "-f/-fix" option as suggested by Stack;
2) Added remarks about "assigns" command in the help text, as pointed out by Daisuke;

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -374,6 +359,8 @@ private static String getCommandUsage() {
     writer.println();
     usageBypass(writer);
     writer.println();
+    usagetExtraRegionsInMeta(writer);
```
daisukebe: Typo: `usageExtraRegionsInMeta`?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -490,6 +477,33 @@ private static void usageReplication(PrintWriter writer) {
     writer.println("   purge if '--fix'.");
   }
 
+  private static void usagetExtraRegionsInMeta(PrintWriter writer) {
```
daisukebe: Same typo: `usageExtraRegionsInMeta`

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
```diff
@@ -33,19 +38,21 @@
...
 import org.apache.hadoop.hbase.util.FSUtils;
 
+import org.apache.hadoop.hbase.util.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * This class implements the inner works required for check and recover regions that wrongly
- * went missing in META.
+ * This class implements the inner works required for checking and recover regions that wrongly
```
joshelser: nit: recovering (to match "checking")

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
```diff
@@ -70,61 +77,207 @@ public FsRegionsMetaRecoverer(Configuration configuration) throws IOException {
...
-        LOG.warn("Can't get missing regions from meta", e);
-      }
+      });
+      return missingRegions;
+    });
+  }
+
+  List<RegionInfo> findExtraRegionsInMETA(String table) throws IOException {
+    InternalMetaChecker<RegionInfo> extraChecker = new InternalMetaChecker<>();
+    return extraChecker.checkRegionsInMETA(table, (regions,dirs) -> {
```
joshelser: Ultimately this is the same logic as `findMissingRegionsInMETA`.

For types `A`, `B`, and `C`: take a `Collection<A>`, `Collection<B>`, a function `f(A) -> C` and `f(B) -> C`. We want to find all elements in `A` that don't exist in `B`. In Set theory, this is the difference: https://en.wikipedia.org/wiki/Complement_(set_theory)#Relative_complement

For this code, we have `Path` (A) and `RegionInfo` (B) and we want to transform them into a common type `String` (C) which is the encoded region name.

Pseudocode-y
`setComplement(Set<RegionInfo> regionsInMeta, Set<Path> regionsInHdfs, Function<RegionInfo,String> regionInfoToName, Function<Path, String> pathToRegionName)`
`setComplement(Set<Path> regionsInHdfs, Set<RegionInfo> regionsInMeta, Function<Path,String> ..., Function<RegionInfo,String> ..)`

If it's not terribly difficult, could the underlying logic be consolidated?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -191,76 +189,63 @@ int setRegionState(ClusterConnection connection, String region,
...
+      new FsRegionsMetaRecoverer(this.conf)) {
+      List<String> namespacesTables = formatNameSpaceTableParam(commandLine.getArgs());
+      Map<TableName, List<RegionInfo>> reportMap =
+        fsRegionsMetaRecoverer.reportTablesExtraRegions(namespacesTables);
+      final List<String> toFix = new ArrayList<>();
+      reportMap.entrySet().forEach( e -> {
+          result.put(e.getKey(),
+            e.getValue().stream().map(r->r.getEncodedName()).collect(Collectors.toList()));
+          if(fix && e.getValue().size()>0){
+            toFix.add(e.getKey().getNameWithNamespaceInclAsString());
```
joshelser: Why convert `TableName` to a `String`? It seems like when you end up using this String, it would be easier (less error-prone) to have the `TableName`.

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
```diff
@@ -70,61 +77,207 @@ public FsRegionsMetaRecoverer(Configuration configuration) throws IOException {
...
+    Map<TableName,List<T>> reportTablesRegions(final List<String> namespacesOrTables,
+      ExecFunction<List<T>, String> checkingFunction) throws IOException {
+      final Map<TableName,List<T>> result = new HashMap<>();
+      List<TableName> tableNames = MetaTableAccessor.
+        getTableStates(FsRegionsMetaRecoverer.this.conn).keySet().stream()
+          .filter(tableName -> {
+            if(namespacesOrTables==null || namespacesOrTables.isEmpty()){
+              return true;
+            } else {
+              Optional<String> findings = namespacesOrTables.stream().filter(
```
joshelser: This is the other half: if we have a `TableName` here, can we simplify this check?

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
```diff
@@ -70,61 +77,207 @@ public FsRegionsMetaRecoverer(Configuration configuration) throws IOException {
...
+              //we want to allow potential running threads to finish, so we collect execution
+              //errors and show those later
+              LOG.debug("Caught execution error: ", e);
+              executionErrors.add(e);
+            }
+          }
+        }
+      } catch (IOException | InterruptedException e) {
+        LOG.error("ERROR executing thread: ", e);
+        throw new IOException(e);
```
joshelser: Unnecessary wrap with another `IOE`.

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
```diff
@@ -70,61 +77,207 @@ public FsRegionsMetaRecoverer(Configuration configuration) throws IOException {
...
+  }
+
+  @FunctionalInterface
+  interface ExecFunction<T, NamespaceOrTable> {
+    T execute(NamespaceOrTable name) throws IOException;
+  }
+
+  public static void main(String[] args){
+    System.out.println("hi");
+  }
```
joshelser: Remove :)

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
wchevreuil: Coming on next commit.

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
wchevreuil: Yeah, I guess we could basically move the whole loop to InternalMetaChecker. Let me work on that and try get it improved in the next commit.

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
wchevreuil: Coming on next commit.

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
wchevreuil: Oops.

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
wchevreuil: Problem is that I'm allowing _namespacesOrTables_ value to be a namespace, in which case, all tables under the given namespace should be checked. If we make it a _TableName_ type, I guess we won't be able to pass a namespace as value.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Will be fixed on next commit.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: Will be fixed on next commit.

wchevreuil:
Ping @joshelser @saintstack @daisukebe , any further thoughts on this one?

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
joshelser: This is nice!

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
joshelser: Answered elsewhere -- string is used to send both table name or namespace

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
```diff
@@ -70,61 +78,203 @@ public FsRegionsMetaRecoverer(Configuration configuration) throws IOException {
...
+                  LOG.debug("running thread for {}", tableName.getNameWithNamespaceInclAsString());
+                  return execFunction.execute(report.get(tableName));
+                }
+              }));
+            } else {
+              LOG.warn("Table {} does not exist! Skipping...",
+                tableName.getNameWithNamespaceInclAsString());
+            }
+          }
+          for(Future<List<String>> f : futures){
```
joshelser: Instead of calling `future.get()` here, what about `future.isDone()` and sending the `Future<List<String>>` back to the caller? It would simplify the logic here and still let you shut down the `ExecutorService`.

Your method return type could just be `Map<String,Future<List<String>>>`, mapping "namespaceOrTable" to a `Future` with the exception or result of computation.

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
```diff
@@ -70,61 +78,203 @@ public FsRegionsMetaRecoverer(Configuration configuration) throws IOException {
...
+
+  @FunctionalInterface
+  interface ExecFunction<T, NamespaceOrTable> {
+    T execute(NamespaceOrTable name) throws IOException;
+  }
+
+  public class ListUtils<T1, T2> {
+    public List<T1> complement(List<T1> list1, List<T2> list2,
+        Function<T2, String> convertFromBase,
+        Function<T1, String> convertFromComparing) throws IOException {
```
joshelser: Weird to have the order of T1 and T2 reversed (1 & 4, then 2 & 3). What about `List<T1>, List<T2>, Function<T1,String>, Function<T2,String>`?

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
```diff
@@ -70,61 +78,203 @@ public FsRegionsMetaRecoverer(Configuration configuration) throws IOException {
...
+      final List<T1> extraRegions = new ArrayList<>();
+      HashSet<String> baseSet = list2.stream().map(info ->
+        convertFromBase.apply(info)).collect(Collectors.toCollection(HashSet::new));
```
joshelser: `convertFromBase` and `convertFromComparing` isn't clear to me what they're doing. Maybe just name them according to the type, e.g. `convertT1` and `convertT2`, since the types are opaque anyways?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -191,76 +189,63 @@ int setRegionState(ClusterConnection connection, String region,
...
-    }
     Pair<List<String>, List<ExecutionException>> result = new Pair<>();
-    result.setFirst(readdedRegionNames);
-    result.setSecond(executionErrors);
+    try (final FsRegionsMetaRecoverer fsRegionsMetaRecoverer =
+      new FsRegionsMetaRecoverer(this.conf)) {
+      result = fsRegionsMetaRecoverer.addMissingRegionsInMetaForTables(
+        formatNameSpaceTableParam(nameSpaceOrTable));
+    } catch (IOException e) {
+      LOG.error("Error adding missing regions: ", e);
```
joshelser: Include `nameSpaceOrTable` in the message?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -843,22 +868,34 @@ private static String toString(List<?> things) {
   }
 
   private String formatMissingRegionsInMetaReport(Map<TableName,List<Path>> report) {
+    Function<Path,String> resolver = r -> r.getName();
+    String message = "Missing Regions for each table:\n\t";
+    return formatReportMessage(message, new HashMap<>(report), resolver);
```
joshelser: Just cast this instead of creating a brand new HashMap?

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBaseFsck.java
```diff
@@ -288,7 +288,7 @@
   /***********
    * Options
    ***********/
-  private static boolean details = false; // do we display the full report
+  private static boolean details = false; // do we display the full execute
```
joshelser: Unintentional change of `s/report/execute/` here and below?

hbase-hbck2/README.md
```diff
@@ -148,6 +148,29 @@ Command:
...
+   file system. Extra regions would get deleted from Meta
+   if passed the --fix option.
+   NOTE: Before deciding on use the "--fix" option, it's worth check if
+   reported extra regions are overlapping with existing valid regions.
+   If so, then "extraRegionsInMeta --fix" is indeed the optimal solution.
+   Otherwise, "assigns" command is the simpler solution, as it recreates
+   regions dirs in the filesystem, if not existing.
+   An example triggering extra regions report for tables 'table_1'
+   and 'table_2', under default namespace:
+     $ HBCK2 extraRegionsInMeta default:table_1 default:table_2
```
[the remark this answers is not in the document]
wchevreuil: Yeah, in the same way as done for the already existing _addFsRegionsMissingInMeta_ and _reportMissingRegionsInMeta_ methods. Since we allow for passing namespaces as well, if we allow for simply passing the table name for tables in default namespace, we would need to do an extra verification for the existence of such table in default namespace, and if no such table found, then fallback to consider it as namespace. And then there's also possibility for a table name in default namespace have same value as a given namespace, which would then prohibit this namespace to be passed. All that made me thought it was better to require explicitly passing the default namespace prefix when operating over specific table.

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
wchevreuil: Maybe I'm not following your thought, but I think doing `future.isDone()` would require extra logic to keep iterating through all futures that are not yet completed, or introduce a latch for control. I think it would still be simpler to keep using `future.get()`, but together with your suggestion for returning `Map<String,Future<List<String>>>`. Let me give it a try.

hbase-hbck2/README.md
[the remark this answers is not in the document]
joshelser: Ok! Just double checking.

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
joshelser: Yeah, `f.get()` is going to be a blocking call, so you would want to have some kind of back-off/sleep/recheck loop here instead (if you choose to take my suggestion ;)).

Something like:
```
ArrayList<Future<List<String>>> completedFutures = new ArrayList<>();
Future<List<String>> current = null;
while (!futures.isEmpty()) {
  Iterator<Future<List<String>>> iter = futures.iterator();
  while (iter.hasNext()) {
    current = iter.next();
    if (current.isDone()) {
      iter.remove();
      completedFutures.add(current);
    }
  }
  if (!futures.isEmpty()) {
    Thread.sleep(2000);
  }
}
```

wchevreuil: Ok, changed to return a `List<Future<List<String>>>`. I realised `future.get()` wouldn't make sense here, I would still need to call it in the caller to actually get results and potential errors.

wchevreuil: Sorry @joshelser , had missed your last comment while I was working on my last change. Let me know what you think on this last one.

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
wchevreuil: Indeed confusing, even I had troubled to figure it out after not touching this code for a month. Changing this and the params order mentioned on above commented on next commit.

hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBaseFsck.java
wchevreuil: Definitely yeah. Reverting this.

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
joshelser: No worries! My suggestion would retain the semantics that you have already (where this method is blocking on completion). You could also just make this return fast, leaving  the author to wait on f.get(). But yeah, leave it up to the caller to decide what is done with errors, IMO (e.g. maybe we find later that HBCK2 should auto-retry a number of times?)

I forget the semantics of `ExecutorService.shutdown()` -- I think that might just reject any new submissions and shut down once everything currently running is done (not creating a "how do I make sure the ES is closed" problem for you).

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
wchevreuil: We don't have how to know for which of the passed namespace or table we are really erroring here, so ain't sure adding the list of all passed namespaces/tables would add much value here.

wchevreuil:
Addressed your suggestions on last commit, @joshelser , am trying to figure out how to make this PR applicable to master now, given all the conflicts. Tried a rebase, but it didn't work, apparently.

wchevreuil:
Looks like my last rebase worked. Waiting on the build results.

wchevreuil:
Ok, only remaining checkstyle issue complains about length of HBCK2.doCommandLine, which unless we do a full re-design of HBCK2 to implement the Command design pattern, I don't think it's addressable, as we will be constantly adding more options there.

wchevreuil:
Signed-off-by: Josh Elser <elserj@apache.org>

joshelser:
🎉

PR #47, opened 2019-12-13
HBASE-23180 hbck2 testing tool

jatsakthi:
This adds a new tool that spins up a hbase on hadoop minicluster and
mimicks actions of hbck2 to verify it's functionalities

### DIFF
--- a/dev-support/testingtool/hbck2_test.sh
+++ b/dev-support/testingtool/hbck2_test.sh
@@ -1,0 +1,353 @@
+#!/usr/bin/env bash
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# This the main control script that takes care of following things:
+#     1. Sets up a hadoop minicluster
+#     2. Installs HBase on the minicluster
+#     3. Creates a table & region data using LTT
+#     4. Invokes all the action scripts present in dev-support/testingtool/scripts/
+
+set -e
+function usage {
+  echo "Usage: ${0} [options] /path/to/component/bin-install /path/to/hadoop/executable /path/to/hadoop/hadoop-yarn-server-tests-tests.jar /path/to/hadoop/hadoop-mapreduce-client-jobclient-tests.jar /path/to/mapred/executable"
+  echo ""
+  echo "    --zookeeper-data /path/to/use                                     Where the embedded zookeeper instance should write its data."
+  echo "                                                                      defaults to 'zk-data' in the working-dir."
+  echo "    --working-dir /path/to/use                                        Path for writing configs and logs. must exist."
+  echo "                                                                      defaults to making a directory via mktemp."
+  echo "    --hadoop-client-classpath /path/to/some.jar:/path/to/another.jar  classpath for hadoop jars."
+  echo "                                                                      defaults to 'hadoop classpath'"
+  echo "    --hbase-client-install /path/to/unpacked/client/tarball           if given we'll look here for hbase client jars instead of the bin-install"
+  echo "    --force-data-clean                                                Delete all data in HDFS and ZK prior to starting up hbase"
+  echo "    --single-process                                                  Run as single process instead of pseudo-distributed"
+  echo "    --hbck2 /path/to/use                                              Path for the hbck2 jar"
+  echo ""
+  exit 1
+}
+# if no args specified, show usage
+if [ $# -lt 5 ]; then
+  usage
+fi
+
+# Get arguments
+declare component_install
+declare hadoop_exec
+declare working_dir
+declare zk_data_dir
+declare clean
+declare distributed="true"
+declare hadoop_jars
+declare hbase_client
+while [ $# -gt 0 ]
+do
+  case "$1" in
+    --working-dir) shift; working_dir=$1; shift;;
+    --force-data-clean) shift; clean="true";;
+    --zookeeper-data) shift; zk_data_dir=$1; shift;;
+    --single-process) shift; distributed="false";;
+    --hadoop-client-classpath) shift; hadoop_jars="$1"; shift;;
+    --hbase-client-install) shift; hbase_client="$1"; shift;;
+    --hbck2) shift; hbck2_jar_path="$1"; shift;;
+    --) shift; break;;
+    -*) usage ;;
+    *)  break;;  # terminate while loop
+  esac
+done
+
+# should still have where component checkout is.
+if [ $# -lt 5 ]; then
+  usage
+fi
+component_install="$(cd "$(dirname "$1")"; pwd)/$(basename "$1")"
+hadoop_exec="$(cd "$(dirname "$2")"; pwd)/$(basename "$2")"
+yarn_server_tests_test_jar="$(cd "$(dirname "$3")"; pwd)/$(basename "$3")"
+mapred_jobclient_test_jar="$(cd "$(dirname "$4")"; pwd)/$(basename "$4")"
+mapred_exec="$(cd "$(dirname "$5")"; pwd)/$(basename "$5")"
+
+if [ ! -x "${hadoop_exec}" ]; then
+  echo "hadoop cli does not appear to be executable." >&2
+  exit 1
+fi
+
+if [ ! -x "${mapred_exec}" ]; then
+  echo "mapred cli does not appear to be executable." >&2
+  exit 1
+fi
+
+if [ ! -d "${component_install}" ]; then
+  echo "Path to HBase binary install should be a directory." >&2
+  exit 1
+fi
+
+if [ ! -f "${yarn_server_tests_test_jar}" ]; then
+  echo "Specified YARN server tests test jar is not a file." >&2
+  exit 1
+fi
+
+if [ ! -f "${mapred_jobclient_test_jar}" ]; then
+  echo "Specified MapReduce jobclient test jar is not a file." >&2
+  exit 1
+fi
+
+if [ -z "${working_dir}" ]; then
+  if ! working_dir="$(mktemp -d -t hbck2-test)" ; then
+    echo "Failed to create temporary working directory. Please specify via --working-dir" >&2
+    exit 1
+  fi
+else
+  # absolutes please
+  working_dir="$(cd "$(dirname "${working_dir}")"; pwd)/$(basename "${working_dir}")"
+  if [ ! -d "${working_dir}" ]; then
+    echo "passed working directory '${working_dir}' must already exist." >&2
+    exit 1
+  fi
+fi
+
+if [ -z "${zk_data_dir}" ]; then
+  zk_data_dir="${working_dir}/zk-data"
+  mkdir "${zk_data_dir}"
+else
+  # absolutes please
+  zk_data_dir="$(cd "$(dirname "${zk_data_dir}")"; pwd)/$(basename "${zk_data_dir}")"
+  if [ ! -d "${zk_data_dir}" ]; then
+    echo "passed directory for zk '${zk_data_dir}' must already exist."
+    exit 1
+  fi
+fi
+
+if [ -z "${hbase_client}" ]; then
+  hbase_client="${component_install}"
+else
+  echo "Using HBase client-side artifact"
+  # absolutes please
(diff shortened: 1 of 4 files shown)


dev-support/testingtool/hbck2_test.sh
```diff
@@ -0,0 +1,354 @@
...
+if [ "${hadoop_version%.*.*}" -gt 2 ]; then
+  "${mapred_exec}" minicluster -format -writeConfig "${working_dir}/hbase-conf/core-site.xml" -writeDetails "${working_dir}/hadoop_cluster_info.json" >"${working_dir}/hadoop_cluster_command.out" 2>"${working_dir}/hadoop_cluster_command.err" &
+else
+  HADOOP_CLASSPATH="${yarn_server_tests_test_jar}" "${hadoop_exec}" jar "${mapred_jobclient_test_jar}" minicluster -format -writeConfig "${working_dir}/hbase-conf/core-site.xml" -writeDetails "${working_dir}/hadoop_cluster_info.json" >"${working_dir}/hadoop_cluster_command.out" 2>"${working_dir}/hadoop_cluster_command.err" &
+fi
+
+echo "$!" > "${working_dir}/hadoop.pid"
+
+sleep_time=2
+until [ -s "${working_dir}/hbase-conf/core-site.xml" ]; do
```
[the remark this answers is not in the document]
jatsakthi: One way to check if the minicluster is up or not is by searching for "INFO mapreduce.MiniHadoopClusterManager: Started MiniMRCluster" in the logs. After the minicluster is started, the configuration details are written in this file so I think even checking the presence of this file would mean the minicluster is already started because it's the next step after logging that line.

dev-support/testingtool/run_hbck2_test.sh
```diff
@@ -0,0 +1,45 @@
...
+SELF=$(cd "$(dirname "$0")" && pwd)
+echo "SELF is at: ${SELF}"
+
+(cd "${SELF}/../../"
+mvn clean package assembly:single -DskipTests
+hbase_hbck2_jar=$(ls hbase-hbck2/*target/hbase-hbck2-*)
+
+# Replace the passed parameters in here [starting from --working-dir] with appropriate values
+./dev-support/testingtool/hbck2_test.sh --hbck2 "${SELF}/../../${hbase_hbck2_jar}" \
+--single-process --working-dir /Users/sakthi/test/hbck2-testing-tool/output/hbck2 \
```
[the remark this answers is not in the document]
jatsakthi: Sure

hbase-hbck2/src/main/java/org/apache/hbase/HBCKActions.java
```diff
@@ -0,0 +1,107 @@
...
+      String tableName = args[1];
+      tool.deleteRegionFromMeta(tableName);
+    }
+    else {
+      System.out.println("ERROR: Unknown option passed.");
+      printUsageAndExit();
+    }
+  }
+
+
```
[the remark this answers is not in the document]
jatsakthi: done

jatsakthi:
Thanks for your comments @petersomogyi . I have updated the PR.

jatsakthi:
@saintstack , @busbey , @wchevreuil can I get any of your reviews as well here?

jatsakthi:
@petersomogyi can I get your +1 here?

dev-support/testingtool/hbck2_test.sh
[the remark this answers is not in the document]
busbey: yes, if you tell Hadoop to write out the configs as a part of the cli args then this file is written out to indicate it is ready for clients.

dev-support/testingtool/hbck2_test.sh
```diff
@@ -0,0 +1,354 @@
...
+while [ $# -gt 0 ]
+do
+  case "$1" in
+    --working-dir) shift; working_dir=$1; shift;;
+    --force-data-clean) shift; clean="true";;
+    --zookeeper-data) shift; zk_data_dir=$1; shift;;
+    --single-process) shift; distributed="false";;
+    --hadoop-client-classpath) shift; hadoop_jars="$1"; shift;;
+    --hbase-client-install) shift; hbase_client="$1"; shift;;
+    --hbck2) shift; hbck2_jar_path="$1"; shift;;
```
busbey: this needs to be declared above before we write to it.

dev-support/testingtool/hbck2_test.sh
```diff
@@ -0,0 +1,354 @@
...
+
+if [ -z "${hadoop_jars}" ]; then
+  echo "Hadoop client jars not given; getting them from 'hadoop classpath'."
+  hadoop_jars=$("${hadoop_exec}" --config "${working_dir}/hbase-conf/" classpath)
+fi
+
+echo "Running all the action scripts from ${SELF}/scripts/..."
+
+echo "Invoking deleteRegionFromMeta.sh..."
+"$SELF/scripts/deleteRegionInMeta.sh" "${hbase_client}" "${working_dir}" "${component_install}" "${hbck2_jar_path}" "${hadoop_jars}" "${hbase_version}" default:testtable
```
busbey: We pass `hbck2_jar_path` here to the underlying script without ever verifying it. We should have similar handling as we do for e.g. `hbase_client` where we make sure we convert it to an absolute path and that it's a file.

if there's nothing we can default it to, then it should be a required parameter ala `yarn_server_tests_test_jar` rather than an optional parameter.

dev-support/testingtool/run_hbck2_test.sh
```diff
@@ -0,0 +1,45 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# This script does following things:
+#     1. Do a mvn package to generate the hbck2 jar
+#     2. Invoke dev-support/testingtool/hbck2_test.sh with the above generated jar and other necessary parameters
+
+# HOW TO RUN:
+#     1. Replace the passed parameters at hbck2_test.sh invocation line with appropriate values
```
busbey: I'm not sure I understand the intended purpose of this script, since it has to be edited before it can be run. Is it an example?

hbase-hbck2/src/main/java/org/apache/hbase/HBCKActions.java
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hbase;
+
+import static java.lang.System.exit;
```
busbey: nit: we only use this once, just call System.exit there.

hbase-hbck2/src/main/java/org/apache/hbase/HBCKActions.java
```diff
@@ -0,0 +1,107 @@
...
+
+  public HBCKActions(Configuration conf) {
+    this.conf = conf;
+  }
+
+  public static void main(String[] args) throws InterruptedException, IOException {
+    Configuration conf = HBaseConfiguration.create();
+    HBCKActions tool = new HBCKActions(conf);
+
+    if (args.length < 1) {
```
busbey: should add a `|| args[0].equals("help")` here so that running with help doesn't include the ERROR line about  an unknown option.

hbase-hbck2/src/main/java/org/apache/hbase/HBCKActions.java
```diff
@@ -0,0 +1,107 @@
...
+      printUsageAndExit();
+    }
+    if (args[0].equals("deleteRegionFromMeta")) {
+      if (args.length != 2) {
+        System.out.println("ERROR: Number of parameters not correct");
```
busbey: nit: System.err

hbase-hbck2/src/main/java/org/apache/hbase/HBCKActions.java
```diff
@@ -0,0 +1,107 @@
...
+        printUsageAndExit();
+      }
+      String tableName = args[1];
+      tool.deleteRegionFromMeta(tableName);
+    }
+    else {
+      System.out.println("ERROR: Unknown option passed.");
```
busbey: nit: System.err

hbase-hbck2/src/main/java/org/apache/hbase/HBCKActions.java
```diff
@@ -0,0 +1,107 @@
...
+      printUsageAndExit();
+    }
+  }
+
+  private static void printUsageAndExit() {
+    System.out.println("hbckActions <options>");
```
busbey: nit: System.err

dev-support/testingtool/scripts/deleteRegionInMeta.sh
```diff
@@ -0,0 +1,119 @@
...
+fi
+
+# Carry forward the arguments from the control script
+declare hbase_client=$1
+declare working_dir=$2
+declare component_install=$3
+declare hbck2_jar_path=$4
+declare hadoop_jars=$5
+declare hbase_version=$6
+# The table name to operate with
```
busbey: nit: unneeded comment

dev-support/testingtool/scripts/deleteRegionInMeta.sh
```diff
@@ -0,0 +1,119 @@
...
+do
+  printf '\tretry waiting for hbase to restart.\n'
+  sleep "${sleep_time}"
+  sleep_time="$((sleep_time*2))"
+done
+
+region_id=$(echo "$assigns_output" | cut -d ' ' -f 2)
+echo "========= Running hbck2's assigns command."
+HBASE_CLASSPATH_PREFIX="${hadoop_jars}" HBASE_CONF_DIR="${working_dir}/hbase-conf/" redirect_and_run \
+    "${working_dir}/hbck2_assignsRegion" "${component_install}/bin/hbase" hbck -j "${hbck2_jar_path}" assigns "$region_id"
```
busbey: a question rather than a concern, do we need the hbck2 assigns here? would a normal assign fail?

dev-support/testingtool/scripts/deleteRegionInMeta.sh
```diff
@@ -0,0 +1,119 @@
...
+
+sleep 5
+echo "========= Verifying row count after assign."
+assign_rowcount=$(echo "count '${table_name}'" | "${hbase_client}/bin/hbase" --config "${working_dir}/hbase-conf/" shell --noninteractive 2>/dev/null | tail -n 1)
+echo "Found ${assign_rowcount} rows..."
+if [ ! "${assign_rowcount}" -eq 1000 ]; then
+  echo "ERROR: Instead of finding 1000 rows, we found ${assign_rowcount}."
+  exit 2
+else
+  echo "SUCCESS: deleteRegionInMeta succedeed"
```
busbey: nit: "SUCCESS: deleteRegionInMeta recovery process succeeded"

dev-support/testingtool/hbck2_test.sh
jatsakthi: Sure. Good catch Sean. I prefer making it a required parameter.

dev-support/testingtool/run_hbck2_test.sh
jatsakthi: Yes, it's more of an example script as of now for someone to use if just by making changes to the parameters that are being passed to hbck2_test.sh. The bigger picture was then when a Jenkins job would be created to automate this testing on a nightly basis, we would just call run_hbck2_test.sh from the Jenkins script. (+ so that run_hbck2_test.sh would take care of compiling and generating a fresh jar every time and pass it to hbck2_test.sh.)

dev-support/testingtool/scripts/deleteRegionInMeta.sh
jatsakthi: A normal assign would also succeed. I guess it's just choosing either of the available options. My preference is hbck2's option as we are trying to verify hbck2's proper functionality through this tool. What do you think Sean?

hbase-hbck2/src/main/java/org/apache/hbase/HBCKActions.java
jatsakthi: Sure

hbase-hbck2/src/main/java/org/apache/hbase/HBCKActions.java
jatsakthi: Took care of these as well.

jatsakthi:
Thanks for your reviews @busbey . I have addressed them and updated the pr.

dev-support/testingtool/scripts/deleteRegionInMeta.sh
busbey: Makes sense. Make a note in a comment so it's clear to future readers? Noting that this test of `assigns` specifically doesn't cover a case where we expect the hbck2 command to work when a normal assign doesn't.

jatsakthi: Sure

jatsakthi:
How does the PR look now, @busbey ?

busbey:
Looks great! Thanks for getting this done.

jatsakthi:
Thanks for your reviews, Peter & Sean. Let me get this in.

PR #52, opened 2020-01-21
HBASE-23714 Updated to version 22 of the Apache parent POM

HorizonNet:


pom.xml
```diff
@@ -236,7 +224,6 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
```
petersomogyi: Is this still required?

HorizonNet: Shouldn't be, as it is the default as the group Id. Left it this way, as we usually do it. Should we start to remove the defaults?

petersomogyi: Bummer, I selected a different line for comment.
I meant the whole `org.apache.maven.plugins:maven-surefire-report-plugin` plugin since there isn't anything configured for it and it is already defined in the ASF pom.

HorizonNet: Checked it and is not required. Removed it.

HorizonNet:
retest build

PR #53, opened 2020-02-04
HBASE-23791 [operator tools] Remove reference to I.A. Private interfa…

wchevreuil:
…ce MetaTableAccessor

### DIFF
--- a/hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
+++ b/hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
@@ -36,7 +36,6 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hbase.HConstants;
-import org.apache.hadoop.hbase.MetaTableAccessor;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.Admin;
 import org.apache.hadoop.hbase.client.Connection;
@@ -112,7 +111,7 @@ public class FsRegionsMetaRecoverer implements Closeable {
 
   void putRegionInfoFromHdfsInMeta(Path region) throws IOException {
     RegionInfo info = HRegionFileSystem.loadRegionInfoFileContent(fs, region);
-    MetaTableAccessor.addRegionToMeta(conn, info);
+    HBCKMetaTableAccessor.addRegionToMeta(conn, info);
   }
 
   List<String> addMissingRegionsInMeta(List<Path> regionsPath) throws IOException {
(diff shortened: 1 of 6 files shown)


wchevreuil:
New commit addressing checkstyle issues for HBCKMetaTableAccessor.

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
```diff
@@ -135,10 +134,16 @@ void putRegionInfoFromHdfsInMeta(Path region) throws IOException {
...
-      }, nameSpaceOrTable);
+      return extraChecker.processRegionsMetaCleanup(this::reportTablesExtraRegions, regions ->
+        regions.stream().map(r -> {
+          try {
+            HBCKMetaTableAccessor.deleteRegionInfo(conn, r);
+            return r.getEncodedName();
+          } catch(IOException e){
+            LOG.error("Failed to delete region: {}", r.getEncodedName());
+            return r.getEncodedName() + " (failed)";
+          }
```
joshelser: This catch seems to be changing the semantics of this method. Instead of the `Future` having an exception thrown, it will always return a `List<String>`. This is a smell as we have to determine success/failure based on the values in that list.

I assume this is an inadvertent introduction with the Function being passed to map not allowing an exception to be thrown?

hbase-hbck2/src/main/java/org/apache/hbase/HBCKMetaTableAccessor.java
```diff
@@ -138,4 +182,550 @@ private static void deleteFromMetaTable(final Connection connection, final Delet
...
+        scan.withStartRow(Bytes.toBytes(startRow));
+        scan.withStopRow(Bytes.toBytes(stopRow));
+      },
+      r -> {
+        Cell cell = r.getColumnLatestCell(CATALOG_FAMILY, REGIONINFO_QUALIFIER);
+        if(cell != null) {
+          RegionInfo info = RegionInfo
+            .parseFromOrNull(cell.getValueArray(), cell.getValueOffset(), cell.getValueLength());
+          return info;
+        } else {
```
joshelser: Nit: can just be `if () { } return null;`

hbase-hbck2/src/main/java/org/apache/hbase/FsRegionsMetaRecoverer.java
wchevreuil: Indeed. Was initially trying to save on the amount of code added/copied from MetaTableAccessor to HBCKMetaTableAccessor, and didn't think it was necessary to define yet another functional interface just to overcome one of most annoying design flaws of java lambda interfaces.

wchevreuil: Ok, just moved the iteration portion to a separate method. Two thoughts I actually had now:
1) I'm deleting regions one by one, meaning, a separate `Table.delete` call for each region passed. We don't expect the number of extra regions to be so high that it would cause major performance impacts, but we could make it more optimal.
2) Whilst last change gets in sync with previous semantic for the method, I wonder if we would rather not fail the entire delete of regions for a given ns/table.

hbase-hbck2/src/main/java/org/apache/hbase/HBCKMetaTableAccessor.java
wchevreuil: Fixed on last commit.

wchevreuil:
> Looks OK, generally. Do you think we should add an enforcer rule to prevent people re-adding MetaAccessor? Hard to say if people will remember not to use it on their own :)
Sounds like a good idea, but maybe it would require a custom rule implementation? MetaTableAccessor is wrapped in hbase-client dependency, which I don't think we can ban entirely in this case.

wchevreuil:
Ping @joshelser : Any further suggestions with this one or are we good to go here?

joshelser:
> Sounds like a good idea, but maybe it would require a custom rule implementation? MetaTableAccessor is wrapped in hbase-client dependency, which I don't think we can ban entirely in this case.

I think we can easily ban MetaTableAccessor with the maven-forbidden-apis plugin: https://github.com/policeman-tools/forbidden-apis/wiki/MavenUsage

We should just be able to add the fully-qualified MetaTableAccessor class name to the signatures file we provide to that plugin. Doesn't need to be done here -- can be done later.

PR #54, opened 2020-03-05
HBASE-23934 [operator tools] Add forbiddennapis plugin to pom.xml so …

wchevreuil:
…that we can permanently ban references to hbase I.A. private classes from hbck2


wchevreuil:
> I'm assuming there's not an easy thing we can do to just flag the IA.Private annotation?

Not that I'm aware of. Any ideas, @busbey ?

busbey:
Nah. We can circle back to it later

PR #55, opened 2020-03-30
HBASE-24039 

richardantal:
HBCK2 feature negotiation to check what commands are supported

### DIFF
--- a/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
+++ b/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
@@ -21,8 +21,10 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
@@ -94,6 +96,15 @@ public class HBCK2 extends Configured implements org.apache.hadoop.util.Tool {
   private static final String SET_REGION_STATE = "setRegionState";
   private static final String SCHEDULE_RECOVERIES = "scheduleRecoveries";
   private static final String FIX_META = "fixMeta";
+  // TODO update this map in case of the name of a method changes in Hbck interface
+  //  in org.apache.hadoop.hbase.client package. Or a new command is added and the hbck command
+  //  does not equals to the method name in Hbck interface.
+  private static final Map<String, List<String> > FUNCTION_NAME_MAP =
+          Collections.unmodifiableMap(new HashMap<String, List<String>>() {{
+              put(SET_TABLE_STATE, Arrays.asList("setTableStateInMeta"));
+              put(BYPASS, Arrays.asList("bypassProcedure"));
+              put(SCHEDULE_RECOVERIES, Arrays.asList("scheduleServerCrashProcedure",
+                      "scheduleServerCrashProcedures")); }});
 
   private static final String ADD_MISSING_REGIONS_IN_META_FOR_TABLES =
     "addFsRegionsMissingInMeta";
@@ -130,6 +141,20 @@ public class HBCK2 extends Configured implements org.apache.hadoop.util.Tool {
             serverVersion + "; needs at least a server that matches or exceeds " +
             Arrays.toString(thresholdVersions));
       }
+    }
+  }
+
+  void checkFunctionSupported(ClusterConnection connection, String cmd) throws IOException {
+    if (skipCheck) {
+      LOG.info("Skipped {} command version check; 'skip' set", cmd);
+      return;
+    }
+    List<Method> methods = Arrays.asList(connection.getHbck().getClass().getDeclaredMethods());
+    List<String> finalCmds = FUNCTION_NAME_MAP.getOrDefault(cmd, Collections.singletonList(cmd));
+    boolean supported = methods.stream().anyMatch(method ->  finalCmds.contains(method.getName()));
+    if (!supported) {
+      throw new UnsupportedOperationException("This HBase cluster does not support command: "
+              + cmd);
     }
   }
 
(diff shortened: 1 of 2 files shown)


petersomogyi:
retest build

petersomogyi:
retest build

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -133,6 +140,19 @@ void checkHBCKSupport(ClusterConnection connection, String cmd, String ... suppo
     }
   }
 
+  void checkFunctionSupported(ClusterConnection connection, String cmd) throws IOException {
```
petersomogyi: With the introduction of this new method can we also remove `checkHBCKSupport`? As I see it is still used in `SET_REGION_STATE`, `FILESYSTEM` and `REPLICATION` commands.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -102,6 +104,11 @@
   private Configuration conf;
   static final String [] MINIMUM_HBCK2_VERSION = {"2.0.3", "2.1.1", "2.2.0", "3.0.0"};
   private boolean skipCheck = false;
+  static final Map<String, List<String> > functionNameMap = new HashMap<String, List<String>>() {{
```
petersomogyi: Can you comment here or to the `checkFunctionSupported` method when this map needs to be updated?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -102,6 +104,11 @@
   private Configuration conf;
   static final String [] MINIMUM_HBCK2_VERSION = {"2.0.3", "2.1.1", "2.2.0", "3.0.0"};
```
petersomogyi: Can this be removed?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
richardantal: The above mentioned functions don't call any method in hbck so I can not do the same check in those cases.
I would stick with the original checkHBCKSupport function call that we used before.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
richardantal: In case of in Hbck interface in org.apache.hadoop.hbase.client package, the name of the method changes. (Can be because of Deprecation or refactoring.) 
I will write it to checkFunctionSupported as well.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
richardantal: checkHBCKSupport is using it and I would get rid of that method yet.

petersomogyi:
@saintstack What you think about this way of compatibility check?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -102,6 +104,14 @@
...
   private boolean skipCheck = false;
+  static final Map<String, List<String> > functionNameMap = new HashMap<String, List<String>>() {{
```
joshelser, line 107: Can you move this Map creation into a static method, and then make it immutable (ala `Collections.unmodifiableMap()`), please?

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -133,6 +143,19 @@ void checkHBCKSupport(ClusterConnection connection, String cmd, String ... suppo
     }
   }
 
+  void checkFunctionSupported(ClusterConnection connection, String cmd) throws IOException {
+    if (skipCheck) {
+      LOG.info("Skipped {} command version check; 'skip' set", cmd);
+      return;
+    }
+    List<Method> methods = Arrays.asList(connection.getHbck().getClass().getDeclaredMethods());
```
joshelser, line 151: This is only checking what is defined in `Hbck` per the local environment. This might differ that the cluster's version.

e.g. a user has their personal computer with HBase 2.2.0 on it, but their cluster they're fixing has 2.2.2 on it.

I think this is OK to start with, but we might want to push some sort of RPC to get the available Hbck methods from the Master itself (when the master, of course, is up) as a future improvement.

saintstack:
@petersomogyi Missed your ping. Sorry.

Regards this approach, I didn't understand why when I saw the JIRA go by but the following rationale makes sense to me: "It is not that easy to decide if the command is supported or not when dealing with back-ported features." I could imagine someone doing loads of backports finding this handy so +1.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
richardantal: I am not an expert in this topic but I think it does not makes sense to run hbck on personal computer to repair a remote HBase. The user should run hbck on the same computer where the problem is. (using ssh for example.) This way the program would get the jar with the correct version.

Anyway we can continue work on a follow up solution where we get methods from the Master itself.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -133,6 +143,19 @@ void checkHBCKSupport(ClusterConnection connection, String cmd, String ... suppo
...
+    List<String> finalCmds = functionNameMap.getOrDefault(cmd, Collections.singletonList(cmd));
+    boolean supported = methods.stream().anyMatch(method ->  finalCmds.contains(method.getName()));
+    if (!supported) {
+      throw new UnsupportedOperationException("Unsupported command" + cmd);
```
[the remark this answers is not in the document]
richardantal: Thanks, I have changed the message.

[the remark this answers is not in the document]
richardantal: I think it is not that easy to remove the version check completely, as we talked about it.

petersomogyi:
retest build

petersomogyi:
The change looks good for me.

Could you fix these checkstyle issues? For some reason the build did not post back to GitHub. https://builds.apache.org/view/H-L/view/HBase/job/PreCommit-HBASE-OPERATOR-TOOLS-Build/169/HTML_20Report/

richardantal:
retest build

petersomogyi:
retest build

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
joshelser: We see this pretty often, actually. Folks will remain on a release line, either upstream releases or downstream re-packagings of those releases.

It would definitely be an improvement to get the discrete list of things that the server supports rather than guessing based on a version string. Like you say, that makes for good follow-on work.

PR #56, opened 2020-03-31
HBASE-23562 [operator tools] Add a RegionsMerge tool that allows for...

BukrosSzabolcs:
…merging multiple adjacent regions until a desired number of regions is reached.

This is a continuation of #46 and was opened because @wchevreuil is temporarily unavailable and I could not add my changes to the original PR.
I made a few changes based on @joshelser 's feedback and a small fix in the merging for cycle's break condition.
@busbey on the open questions, would you think the CLI option a must for this first version?


hbase-tools/pom.xml
```diff
@@ -0,0 +1,202 @@
...
+
+        <dependency>
+            <groupId>org.apache.hbase</groupId>
+            <artifactId>hbase-server</artifactId>
+            <version>${hbase.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hbase</groupId>
+            <artifactId>hbase-shaded-testing-util</artifactId>
```
joshelser, line 66: Should this be a test scope dependency?

hbase-tools/src/main/java/org/apache/hbase/RegionsMerger.java
```diff
@@ -0,0 +1,252 @@
...
+ * HBase maintenance tool for merging regions of a specific table, until a target number of regions
+ * for the table is reached, or no more merges can complete due to limit in resulting merged
+ * region size.
+ */
+public class RegionsMerger extends Configured implements org.apache.hadoop.util.Tool {
+
+  private static final Logger LOG = LoggerFactory.getLogger(RegionsMerger.class.getName());
+  public static final String RESULTING_REGION_UPPER_MARK = "hbase.tools.merge.upper.mark";
+  public static final String SLEEP = "hbase.tools.merge.sleep";
+  public static final String MAX_ROUNDS_IDDLE = "hbase.tools.max.iterations.blocked";
```
joshelser, line 71: nit: IDDLE

hbase-tools/src/main/java/org/apache/hbase/RegionsMerger.java
```diff
@@ -0,0 +1,252 @@
...
+    RowFilter rowFilter = new RowFilter(CompareOperator.EQUAL,
+      new SubstringComparator(tblName+","));
+    SingleColumnValueFilter colFilter = new SingleColumnValueFilter(CATALOG_FAMILY,
+      STATE_QUALIFIER, CompareOperator.EQUAL, Bytes.toBytes("OPEN"));
+    Scan scan = new Scan();
+    FilterList filter = new FilterList(FilterList.Operator.MUST_PASS_ALL);
+    filter.addFilter(rowFilter);
+    filter.addFilter(colFilter);
+    scan.setFilter(filter);
+    ResultScanner rs = metaTbl.getScanner(scan);
```
joshelser, line 122: try/finally to close the `ResultScanner` please.

hbase-tools/src/main/java/org/apache/hbase/RegionsMerger.java
```diff
@@ -0,0 +1,252 @@
...
+    } catch(Exception e){
+      LOG.error("Merging regions failed:", e);
+      return 2;
+    }
+    return 0;
+  }
+
+  public static void main(String [] args) throws Exception {
+    Configuration conf = HBaseConfiguration.create();
+    int errCode = org.apache.hadoop.util.ToolRunner.run(new RegionsMerger(conf), args);
```
joshelser, line 247: nit: is there a reason we're using the fully qualified class name for ToolRunner here? (vs. using an import)

hbase-tools/src/test/java/org/apache/hbase/TestRegionsMerger.java
```diff
@@ -0,0 +1,118 @@
...
+    TEST_UTIL.deleteTable(TABLE_NAME);
+  }
+
+  @Test
+  public void testMergeRegionsCanMergeToTarget() throws Exception {
+    TEST_UTIL.getConfiguration().setInt(RegionsMerger.MAX_ROUNDS_IDDLE, 10);
+    RegionsMerger merger = new RegionsMerger(TEST_UTIL.getConfiguration());
+    merger.mergeRegions(TABLE_NAME.getNameWithNamespaceInclAsString(), 3);
+    List<RegionInfo> result = TEST_UTIL.getAdmin().getRegions(TABLE_NAME);
+    assertEquals(3, result.size());
```
joshelser, line 75: For all of the tests, can you please add a verification step to make sure that we can still read all records in the table?

PR #58, opened 2020-04-21
HBASE-24222 remove FSUtils.checkAccess and replace with FileSystem.access in HBCK

nyl3532016:


hbase-hbck2/src/main/java/org/apache/hbase/hbck1/HBaseFsck.java
```diff
@@ -2450,7 +2450,7 @@ private void checkAndFixTableStates() throws IOException {
     }
   }
 
-  private void preCheckPermission() throws IOException, AccessDeniedException {
+  private void preCheckPermission() throws IOException, AccessControlException {
```
jojochuang, line 2453: In printUsageAndExit() which invokes preCheckPermission(), should it also update the following catch clause to catch AccessControlException instead?
`
    } catch (AccessDeniedException ace) {
`

nyl3532016: printUsageAndExit do not invoke preCheckPermission ? I can not find  the invoke

liuml07: The only place I found was `HBaseFsck.java`, which deals with this the same as generic IOE (simply `exit`)

nyl3532016:
@liuml07 @jojochuang please hep merge this one

liuml07:
> @liuml07 @jojochuang please help merge this one

I don’t have commit access. We can ping @Apache9 to have a look.

jatsakthi:
+1. Thanks for the contribution @nyl3532016 and the review @liuml07 & @jojochuang . Let me merge this one.

PR #62, opened 2020-06-08
HBCK takes one or more files for assigns

clarax:
This is for https://issues.apache.org/jira/browse/HBASE-23927

### DIFF
--- a/hbase-hbck2/README.md
+++ b/hbase-hbck2/README.md
@@ -124,17 +124,20 @@ Command:
    SEE ALSO: reportMissingRegionsInMeta
    SEE ALSO: fixMeta
 
- assigns [OPTIONS] <ENCODED_REGIONNAME>...
+ assigns [OPTIONS] <ENCODED_REGIONNAME/INPUTFILES_FOR_REGIONNAMES>...
    Options:
     -o,--override  override ownership by another procedure
+    -i,--inputFiles  take one or more encoded region names
    A 'raw' assign that can be used even during Master initialization (if
    the -skip flag is specified). Skirts Coprocessors. Pass one or more
    encoded region names. 1588230740 is the hard-coded name for the
    hbase:meta region and de00010733901a05f5a2a3a382e27dd4 is an example of
    what a user-space encoded region name looks like. For example:
-     $ HBCK2 assign 1588230740 de00010733901a05f5a2a3a382e27dd4
+     $ HBCK2 assigns 1588230740 de00010733901a05f5a2a3a382e27dd4
    Returns the pid(s) of the created AssignProcedure(s) or -1 if none.
-
+   If -i or --inputFiles is specified, pass one or more input file names.
+   Each file contains encoded region names, one per line. For example:
+     $ HBCK2 assigns -i fileName1 fileName2
  bypass [OPTIONS] <PID>...
    Options:
     -o,--override   override if procedure is running/stuck
--- a/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
+++ b/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
@@ -90,7 +90,6 @@ public class HBCK2 extends Configured implements org.apache.hadoop.util.Tool {
   static final int EXIT_FAILURE = 1;
   // Commands
   private static final String SET_TABLE_STATE = "setTableState";
-  private static final String ASSIGNS = "assigns";
   private static final String UNASSIGNS = "unassigns";
   private static final String BYPASS = "bypass";
   private static final String FILESYSTEM = "filesystem";
@@ -112,7 +111,10 @@ public class HBCK2 extends Configured implements org.apache.hadoop.util.Tool {
   private static final String ADD_MISSING_REGIONS_IN_META_FOR_TABLES =
     "addFsRegionsMissingInMeta";
   private static final String REPORT_MISSING_REGIONS_IN_META = "reportMissingRegionsInMeta";
+
   static final String EXTRA_REGIONS_IN_META = "extraRegionsInMeta";
+  static final String ASSIGNS = "assigns";
+
   private Configuration conf;
   static final String [] MINIMUM_HBCK2_VERSION = {"2.0.3", "2.1.1", "2.2.0", "3.0.0"};
   private boolean skipCheck = false;
--- a/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
+++ b/hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
@@ -17,6 +17,9 @@
  */
 package org.apache.hbase;
 
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintWriter;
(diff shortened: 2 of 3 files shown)


hbase-hbck2/README.md
```diff
@@ -124,17 +124,20 @@ Command:
...
    A 'raw' assign that can be used even during Master initialization (if
    the -skip flag is specified). Skirts Coprocessors. Pass one or more
    encoded region names. 1588230740 is the hard-coded name for the
    hbase:meta region and de00010733901a05f5a2a3a382e27dd4 is an example of
    what a user-space encoded region name looks like. For example:
-     $ HBCK2 assign 1588230740 de00010733901a05f5a2a3a382e27dd4
+     $ HBCK2 assigns 1588230740 de00010733901a05f5a2a3a382e27dd4
    Returns the pid(s) of the created AssignProcedure(s) or -1 if none.
-
+   If -i or --inputFiles is specified, pass one or more input file names.
```
huaxiangsun, line 138: Are multiple input file too much? Maybe one file is good enough.

huaxiangsun: Read again, the code only handles one input file, we can clean up the comments/help message accordingly.

clarax: I am using the same optional params for regions so that it requires less change of the existing argument parsing logic and documentation. It might be also useful to support more than one file if users have more than one dump from different criteria to read from.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -294,7 +299,29 @@ int setRegionState(ClusterConnection connection, String region,
...
     }
     boolean overrideFlag = commandLine.hasOption(override.getOpt());
-    return hbck.assigns(commandLine.getArgList(), overrideFlag);
+
+    List<String> argList = commandLine.getArgList();
+    if (!commandLine.hasOption(inputFile.getOpt())) {
+      return hbck.assigns(argList, overrideFlag);
+    } else {
+      List<String> assignmentList = new ArrayList<>();
+      for (String filePath : argList) {
```
clarax, line 308: Here, multiple input files are supported.

hbase-hbck2/src/test/java/org/apache/hbase/TestHBCK2.java
```diff
@@ -147,6 +149,27 @@ public void testAssigns() throws IOException {
           LOG.info("RS: {}", rs.toString());
           assertTrue(rs.toString(), rs.isOpened());
         }
+        // test input files
```
huaxiangsun, line 152: Can we have test similar to the following? That will really go through the option processing path.
The output will have a set of pids and we wait for pids to check the regions are opened.
```
  private String testFormatExtraRegionsInMeta(String[] args) throws IOException {
    HBCK2 hbck = new HBCK2(TEST_UTIL.getConfiguration());
    final StringBuilder builder = new StringBuilder();
    PrintStream originalOS = System.out;
    OutputStream testOS = new OutputStream() {
      @Override public void write(int b) throws IOException {
        builder.append((char)b);
      }
    };
    System.setOut(new PrintStream(testOS));
    hbck.run(args);
    System.setOut(originalOS);
    return builder.toString();
```

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
huaxiangsun: Got it, thanks.

hbase-hbck2/README.md
huaxiangsun: Good.

hbase-hbck2/src/test/java/org/apache/hbase/TestHBCK2.java
```diff
@@ -152,17 +152,11 @@ public void testAssigns() throws IOException {
         // test input files
         String testFile = "inputForAssignsTest";
         FileOutputStream output = new FileOutputStream(testFile, false);
-        LOG.info("writing to: {}", testFile);
```
huaxiangsun: Regions in regionStrsArray has been assigned already. These regions need to be unassigned first before testing assigns with file option.

```
List<Long> pids = this.hbck2.unassigns(hbck, regionStrsArray);
        waitOnPids(pids);
```
The only missing part from testRunWithArgs is to get pid output from console and then wait on them. 
Another way is to wait until the active procedure count to 0.
```
MasterServices services = TEST_UTIL.getHBaseCluster().getMaster();

while (!services.getMasterProcedureExecutor().getActiveProcIds().isEmpty()) {
   sleep (200);
}
```

busbey:
this merged without a jira id. @huaxiangsun please revert and apply again with a corrected commit message.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -109,7 +111,10 @@
   private static final String ADD_MISSING_REGIONS_IN_META_FOR_TABLES =
     "addFsRegionsMissingInMeta";
   private static final String REPORT_MISSING_REGIONS_IN_META = "reportMissingRegionsInMeta";
+
   static final String EXTRA_REGIONS_IN_META = "extraRegionsInMeta";
+  static final String ASSIGNS = "assigns";
```
ndimiduk, line 116: nit: I prefer constants used in APIs/CLIs to remain private, and the tests to duplicate the values. That way, it's easier to notice a breaking change when refactoring existing code -- the test that duplicates the value (rather than referring to the constant in the source class) will start breaking.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -294,7 +301,29 @@ int setRegionState(ClusterConnection connection, String region,
...
+      return hbck.assigns(argList, overrideFlag);
+    } else {
+      List<String> assignmentList = new ArrayList<>();
+      for (String filePath : argList) {
+        try {
+          File file = new File(filePath);
+          FileReader fileReader = new FileReader(file);
+          BufferedReader bufferedReader = new BufferedReader(fileReader);
+          String regionName;
+          while ((regionName = bufferedReader.readLine()) != null) {
```
ndimiduk, line 316: doesn't check style complain about this while-loop construct? Oh, I see we have no precommit job for this project :(

hbase-hbck2/src/test/java/org/apache/hbase/TestHBCK2.java
```diff
@@ -127,32 +130,40 @@ public void testAssigns() throws IOException {
...
+        }
+        output.close();
+        String result = testRunWithArgs(new String[] {HBCK2.ASSIGNS, "-i", testFile});
+        Scanner scanner = new Scanner(result).useDelimiter("[\\D]+");
+        pids = new ArrayList<>();
+        while (scanner.hasNext()) {
+            pids.add(scanner.nextLong());
+        }
+        scanner.close();
+        //pids = this.hbck2.assigns(hbck, new String[] {"-i", testFile});
```
ndimiduk, line 164: delete stray comment?

hbase-hbck2/src/test/java/org/apache/hbase/TestHBCK2.java
```diff
@@ -266,6 +277,29 @@ public void testFormatReportMissingInMetaOneMissing() throws IOException {
     assertTrue(result.contains(expectedResult));
   }
 
+  private void unassigns(List<RegionInfo> regions, String[] regionStrsArray) throws IOException {
```
ndimiduk, line 280: nice refactor.

huaxiangsun:
Sorry @busbey, did not notice that jira id is missing. I will revert and ask @clarax to resubmit the patch with correct jira info.

PR #64, opened 2020-06-09
HBASE-23927 recommit the fix with updates for PR review feedback

clarax:


hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -294,7 +300,30 @@ int setRegionState(ClusterConnection connection, String region,
       return null;
     }
     boolean overrideFlag = commandLine.hasOption(override.getOpt());
-    return hbck.assigns(commandLine.getArgList(), overrideFlag);
+
+    List<String> argList = commandLine.getArgList();
+    if (!commandLine.hasOption(inputFile.getOpt())) {
+      return hbck.assigns(argList, overrideFlag);
+    } else {
```
ndimiduk, line 307: nit: else block is redundant to the early `return` the line prior.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -294,7 +300,30 @@ int setRegionState(ClusterConnection connection, String region,
...
+      List<String> assignmentList = new ArrayList<>();
+      for (String filePath : argList) {
+        try {
+          File file = new File(filePath);
+          FileReader fileReader = new FileReader(file);
+          BufferedReader bufferedReader = new BufferedReader(fileReader);
+          String regionName = bufferedReader.readLine().trim();
+          while (regionName != null) {
```
ndimiduk, line 315: Sorry I wasn't more specific before. How about something like [IOUtils.lineIterator](http://commons.apache.org/proper/commons-io/javadocs/api-release/org/apache/commons/io/IOUtils.html#lineIterator-java.io.InputStream-java.nio.charset.Charset-)? That way there's no infinite loops and no null checking, just a simple for loop with an iterator.

hbase-hbck2/src/main/java/org/apache/hbase/HBCK2.java
```diff
@@ -441,16 +470,20 @@ private static void usageAddFsRegionsMissingInMeta(PrintWriter writer) {
   }
 
   private static void usageAssigns(PrintWriter writer) {
-    writer.println(" " + ASSIGNS + " [OPTIONS] <ENCODED_REGIONNAME>...");
+    writer.println(" " + ASSIGNS + " [OPTIONS] <ENCODED_REGIONNAME/INPUTFILES_FOR_REGIONNAMES>...");
     writer.println("   Options:");
     writer.println("    -o,--override  override ownership by another procedure");
+    writer.println("    -i,--inputFiles  take one or more files of encoded region names");
```
ndimiduk, line 476: yikes! i'm surprised our cli parsing library doesn't handle generating a help message from arguments on our behalf :(

hbase-hbck2/src/test/java/org/apache/hbase/TestHBCK2.java
```diff
@@ -70,6 +73,8 @@
   private static final TableName TABLE_NAME = TableName.valueOf(TestHBCK2.class.getSimpleName());
   private static final TableName REGION_STATES_TABLE_NAME = TableName.
     valueOf(TestHBCK2.class.getSimpleName() + "-REGIONS_STATES");
+  private final static String ASSIGNS = "assigns";
```
ndimiduk, line 76: thanks!

hbase-hbck2/src/test/java/org/apache/hbase/TestHBCK2.java
```diff
@@ -127,32 +132,39 @@ public void testAssigns() throws IOException {
             getRegionStates().getRegionState(ri.getEncodedName());
         LOG.info("RS: {}", rs.toString());
       }
-      List<String> regionStrs =
-          regions.stream().map(RegionInfo::getEncodedName).collect(Collectors.toList());
-      String [] regionStrsArray = regionStrs.toArray(new String[] {});
+      String [] regionStrsArray  =
+          regions.stream().map(RegionInfo::getEncodedName).collect(Collectors.toList())
+                  .toArray(new String[] {});
```
ndimiduk, line 137: You can skip the `collect` and go directly [`toArray`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#toArray-java.util.function.IntFunction-).

PR #65, opened 2020-06-16
HBASE-24571 HBCK2 fix addFsRegionsMissingInMeta to add regions in CLO…

wchevreuil:
…SED state again

This PR fix implementation of HBCKMetaTableAccessor.addRegionToMeta to insert new regions with CLOSED state, and also adds extra checks for this behaviour on both TestHBCKMetaTableAccessor and TestFsRegionsMetaRecoverer UTs.

### DIFF
--- a/hbase-hbck2/src/main/java/org/apache/hbase/HBCKMetaTableAccessor.java
+++ b/hbase-hbck2/src/main/java/org/apache/hbase/HBCKMetaTableAccessor.java
@@ -457,7 +457,7 @@ public final class HBCKMetaTableAccessor {
     return put;
   }
 
-  private static Put addRegionStateToPut(Put put, RegionState.State state) throws IOException {
+  private static void addRegionStateToPut(Put put, RegionState.State state) throws IOException {
     put.add(CellBuilderFactory.create(CellBuilderType.SHALLOW_COPY)
       .setRow(put.getRow())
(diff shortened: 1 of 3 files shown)


wchevreuil:
The failure is not related to this commit. It seems the test added in HBASE-23927 is creating an _inputForAssignsTest_ file under project root dir, which causes asflicense checks to fail: https://builds.apache.org/job/PreCommit-HBASE-OPERATOR-TOOLS-Build/175/artifact/patchprocess/patch-asflicense-problems.txt.

wchevreuil:
retest build

hbase-hbck2/src/main/java/org/apache/hbase/HBCKMetaTableAccessor.java
```diff
@@ -454,6 +457,18 @@ public static Put makePutFromRegionInfo(RegionInfo region, long ts) throws IOExc
     return put;
   }
 
+  private static Put addRegionStateToPut(Put put, RegionState.State state) throws IOException {
```
virajjasani, line 460: This could be `void addRegionStateToPut` since return value is of no use?

wchevreuil: Yes.

PR #66, opened 2020-06-16
HBASE-23927 hbck2 assigns command should accept a file containing a list of region names - amend to put test input file in test data dir 

clarax:


wchevreuil:
retest build

PR #68, opened 2020-06-18
HBASE-23927 hbck2 command should accept one or more files containing …

clarax:
…a list of region names


clarax:
https://github.com/apache/hbase-operator-tools/pull/69
Created a new Jira and PR.

PR #71, opened 2020-07-27
HBASE-24778 [hbase-operator-tools] Merging regions failed when the table is not default namespace

ZhaoBQ:
[HBASE-24778](https://issues.apache.org/jira/browse/HBASE-24778)

### DIFF
--- a/hbase-tools/src/main/java/org/apache/hbase/RegionsMerger.java
+++ b/hbase-tools/src/main/java/org/apache/hbase/RegionsMerger.java
@@ -90,7 +90,7 @@ public class RegionsMerger extends Configured implements org.apache.hadoop.util.
     Path basePath = new Path(conf.get(HConstants.HBASE_DIR));
     basePath = new Path(basePath, "data");
     Path tablePath = new Path(basePath, table.getNamespaceAsString());
-    return new Path(tablePath, table.getNameAsString());
+    return new Path(tablePath, table.getQualifierAsString());
   }
 
   private long sumSizeInFS(Path parentPath) throws IOException {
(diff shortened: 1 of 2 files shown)


ZhaoBQ:
It's just that the getTablePath method is wrong. And I found that FSUtils#getTableDir can be used directly. Do I still need to add UT for non default namespace? @wchevreuil

hbase-tools/src/main/java/org/apache/hbase/RegionsMerger.java
```diff
@@ -88,9 +89,7 @@ public RegionsMerger(Configuration conf) throws IOException {
 
   private Path getTablePath(TableName table){
     Path basePath = new Path(conf.get(HConstants.HBASE_DIR));
-    basePath = new Path(basePath, "data");
-    Path tablePath = new Path(basePath, table.getNamespaceAsString());
-    return new Path(tablePath, table.getNameAsString());
+    return FSUtils.getTableDir(basePath, table);
```
wchevreuil, line 92: FSUtils is IA private, we shouldn't use it in hbase-operator-tools as it makes no compatibility guarantees. See HBASE-24626, HBASE-24482 and HBASE-23791 for issues caused by dependency on IA Private classes. Please revert this to your previous commit. 

Tip: You may want to port this method to hbase-hbck2 HBCKFsUtils class.

ZhaoBQ:
Please review @wchevreuil

ZhaoBQ:
The common part of all test methods is abstracted into mergeRegionsToTarget method. Please review @wchevreuil

PR #73, opened 2020-08-14
HBASE-24882 Migrate hbase-operator-tools testing to ci-hadoop

petersomogyi:


petersomogyi:
Test Pull Request build: https://ci-hadoop.apache.org/job/HBase/job/HBase-Operator-Tools-PreCommit/view/change-requests/job/PR-1/14/

After this is merged the job needs to be updated to use `apache/hbase-operator-tools` repository instead of `petersomogyi/hbase-operator-tools`.

dev-support/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,142 @@
...
+                dir("${WORKDIR}") {
+                    withCredentials([
+                      usernamePassword(
+                        credentialsId: 'apache-hbase-at-github.com',
+                        passwordVariable: 'GITHUB_PASSWORD',
+                        usernameVariable: 'GITHUB_USER'
+                      )]) {
+                        sh label: 'test-patch', script: '''#!/bin/bash -e
+                            hostname -a ; pwd ; ls -la
+                            printenv 2>&1 | sort
```
busbey, line 82: if we have `gather_machine_environment.sh` running why do we need these calls here rather than in there?

busbey, line 82: calling printenv seems dangerous. are we sure there a) aren't credentials in there or b) jenkins console output will filter them if present?

petersomogyi: Jenkins filters out sensitive information. This is the same with other PR builds we have.
```
17:00:32  GITHUB_PASSWORD=****
17:00:32  GITHUB_USER=****
```
I think it won't be filtered if this is printed to a file in `gather_machine_environment.sh` and adding our own logic would be problematic. A contributor could just change gather_machine_environment.sh in a PR not to filter out GitHub user/pass and get the credentials.

petersomogyi: Valid comment, I'll move hostname to gather_machine_environment.sh. I think pwd and ls are not important, the path is already printed out by Jenkins at the beginning of the build along with the worker node name.
Unfortunately, this does not match the hostname: H2 is asf902 but H18 is asf914. 🙁

petersomogyi:
On a different PR I only changed a test file and the first build on Yetus failed but a retrigger succeeded.
The outputs for maven commands are empty.
https://ci-hadoop.apache.org/view/HBase/job/HBase/job/HBase-Operator-Tools-PreCommit/view/change-requests/job/PR-2/ 

@busbey, have you seen something similar before?

busbey:
> On a different PR I only changed a test file and the first build on Yetus failed but a retrigger succeeded.

yeah, this is a long standing bug with jenkins where it does not set default env variables on first run.

